PR #11, opened 2017-03-20
chore: fix coverage results for all packages

JaKXz:
also address issue with istanbul-api package, hopefully.


package.json
```diff
@@ -28,5 +28,16 @@
     "cross-env": "^3.2.3",
     "lerna": "^2.0.0-beta.38-31-g0749ea6",
     "nyc": "^10.2.0-candidate.0"
+  },
+  "nyc": {
+    "exclude": [
+      "**/test/**",
+      "**/dist/**",
+      "**/bin/**"
```
JaKXz: TBPH I'm not a huge fan of this because it seems brittle (i.e. what if we hypothetically wanted to cover something in a `dist/` folder?) but I think the compromise works because of the use case. Let me know what you all think :)

package.json
```diff
@@ -28,5 +28,16 @@
...
+    ],
+    "reporter": [
+      "text",
+      "lcov"
```
JaKXz: Added this for a local verification. I'm not overly attached to it, but `text` and `lcov` are usually the reporters I have on every project... but I am open to removing it for the sake of performance or some other valid reason.

JaKXz:
@bcoe I believe so, but I can remove it in case.

PR #17, opened 2017-03-21
More accurate overall coverage, less configuration

JaKXz:


JaKXz:
@bcoe My guess (say, 50% confidence) at what got things working in #11 for `istanbul-api` was just fixing configuration like this. I don't think `mocha` takes arguments after the `--` like `npm` does.

bcoe:
@JaKXz this is looking good to land?

PR #21, opened 2017-03-23
chore(build): add appveyor config

JaKXz:


JaKXz:
@mattlewis92 FWIW I think I have a legit failure on Windows? It's in `istanbul-lib-sourcemaps/transformer`.

mattlewis92:
@JaKXz I'm not actually a windows user so can't offer much help, I was just fixing a bug for a user of [one of my modules](https://github.com/mattlewis92/karma-coverage-istanbul-reporter)

bcoe:
@JaKXz you are an admin now 👍 might be worth looking at how I've incrementally added windows tests here: https://github.com/conventional-changelog/conventional-changelog/blob/master/appveyor.yml

bcoe:
@JaKXz friendly bump 👍 let me know anything you need help with to get this over the finish line.

JaKXz:
@bcoe know of a way I can quickly debug this on my Mac? Unfortunately the cycles are really slow right now and I don't have access to a windows machine/image. I also don't know why that specific test is _actually_ failing as opposed to just timing out...

bcoe:
@JaKXz https://github.com/bcoe/awesome-cross-platform-nodejs#virtualization

[ievms](https://github.com/xdissent/ievms) is a life saver.

lerna.json
```diff
@@ -3,5 +3,6 @@
   "packages": [
```
bcoe: I think you need to update the lerna version here too, if you're going to upgrade lerna.

JaKXz: Done!

JaKXz:
Thanks for the review @bcoe but I'm really uncomfortable merging this while lerna is still so flaky on windows. Probably not lerna's fault but I'll cc @hzoo if he has any thoughts :)

JaKXz:
@bcoe seems like v2.0.0-rc.3 is a lot more stable :) and I think the `--hoist` flag makes sense in general. What do you think?

bcoe:
@JaKXz great work 👍

PR #24, opened 2017-03-29
Add possibility to filter coverage maps on absolute filename

schutm:
As discussed, previous changes redone in the mono-repo and introduced a parameter/flag in stead of a separate module.

The problem I try to resolve:
I've a 'non-standard' setup, where I build my distribution-packages outside of the source folder. I do this because I write mainly ES6 code, and distribution-packages need to be plain old JavaScript. In addition because AVA is perfectly able to run ES6 tests, but not able to test ES6 modules I've to transpile my ES6 code (using rollup and babel/bublé) to plain JavaScript. The resulting transpiled file will also include all libraries being `import`ed.

The indication source files and line missing coverage works perfectly, using sourcemaps. However in the sourcemaps the original full paths are included. So if I run `nyc` with an include or exclude filter the I only have 2 options: show the coverage of all modules (including external modules, which I didn't intend to be instrumented to be covered), or no modules at all, because all of the sourcemaps absolute paths are outside of the current directory. 

Therefore I introduced these methods, which I intend to use from `nyc` (either by having the `--report` flag taking a different or by introducing a flag (`--no-relativepath`, `--absolutepath` or something similar).

Now writing this all down, a solution might be to first have `nyc` do its work, including generating the report-dir, but no output. Changing to the original source directory again and invoking `nyc --report --report-dir=<report-dir>`. I've to validate this works.

### DIFF
--- a/packages/istanbul-lib-coverage/lib/coverage-map.js
+++ b/packages/istanbul-lib-coverage/lib/coverage-map.js
@@ -59,8 +59,9 @@ CoverageMap.prototype.merge = function (obj) {
 };
 /**
  * filter the coveragemap based on the callback provided
- * @param {Function (filename)} callback - Returns true if the should be
- *  include in the coveragemap. False if it should be removed.
+ * @param {Function (filename)} callback - Returns true if the path
+ *  should be included in the coveragemap. False if it should be
+ *  removed.
  */
 CoverageMap.prototype.filter = function (callback) {
      var that = this;
--- a/packages/istanbul-lib-coverage/lib/coverage-map.js
+++ b/packages/istanbul-lib-coverage/lib/coverage-map.js
@@ -58,6 +58,19 @@ CoverageMap.prototype.merge = function (obj) {
     });
 };
 /**
+ * filter the coveragemap based on the callback provided
+ * @param {Function (filename)} callback - Returns true if the should be
+ *  include in the coveragemap. False if it should be removed.
+ */
+CoverageMap.prototype.filter = function (callback) {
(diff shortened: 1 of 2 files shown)


packages/istanbul-lib-coverage/lib/coverage-map.js
```diff
@@ -58,6 +58,19 @@ CoverageMap.prototype.merge = function (obj) {
     });
 };
 /**
+ * filter the coveragemap based on the callback provided
+ * @param {Function (filename)} callback - Returns true if the should be
```
bcoe: typo, should probably read: "Returns true if the path should be"?

bcoe: I think I'm blanking on how this additional filter option relates to the change in `test-exclude`, mind providing an example of the final functionality you're envisioning in the pull request description.

This is looking great btw, thanks for the contribution 👍

schutm: Fixed the typo. And added additional information in the pull request description.

schutm:
I checked whether I could work around my issue in another way (using `--report-dir`) which didn't work. I misread the help-explanation of this flag, it is no directory containing the raw report data, but the report output itself. In addition I tried to copy the `.nyc-report` directory to the original source directory, but running `nyc --report` in that case still shows all instrumented files. It seems the filtering (to only include files in the current directory) is only performed during instrumentation and not during reporting. Something already indicated by the method name: `shouldInstrument`

PR #32, opened 2017-04-12
tests(istanbul-api): use before for expensive test op

JaKXz:
Extremely minor potential improvement... 

Closes https://github.com/istanbuljs-archived-repos/istanbul-api/pull/20


bcoe:
@JaKXz good thinking 😄

Issue #28, opened 2017-04-06
html reporter not working


Stupidism:
```
fs.js:951
  return binding.readdir(pathModule._makeLong(path), options.encoding);
                 ^

Error: ENOENT: no such file or directory, scandir 'xxxx/node_modules/nyc/node_modules/istanbul-reports/lib/html/assets'
    at Error (native)
    at Object.fs.readdirSync (fs.js:951:18)

```

bcoe:
@Stupidism could you share a project (or a branch of a project) that is exhibiting this issue?

I haven't been able to reproduce on my own projects, which makes me think there must ben an edge-case that's triggering this behavior.

Stupidism:
Wierd, it's gone in my project. It must be an occasional bug.

Stupidism:
It happens again, here is my detailed tests:
## commands
- A: npm run test:cover
- B-G(global): nyc report --reporter=html
- B-L(local): ./node_modules/.bin/nyc report --reporter=html
- C(npm scripts): A && B-G

| install tool        | npm           | yarn  |
| ------------- |:-------------:| -----:|
| run `A` and then `B-G`   | working | working |
| run `A` and then `B-L`      | working      |   working |
| run `A && B-G`      | working      |   working |
| run `A && B-L`      | working      |   error |
| run `C`      | working      |   error |

Stupidism:
Not sure this is a problem with `yarn` or `istanbul`
Possible reasons:
1. `yarn` has a bug
2. `istanbul` used some not-fullly-supported-by-yarn hook of npm.

Stupidism:
@bcoe updates

bcoe:
@Stupidism it sounds like this was an issue with yarn, that may have been resolved; mind testing things out with the latest version?

bcoe:
@Stupidism I dug into this problem a bit further for you, the problem is the order of the rewire and istanbul plugin:

https://github.com/speedskater/babel-plugin-rewire#istanbul

If you swap the order of istanbul, and rewire in @JaKXz's example configuration, things should start working for you.

Stupidism:
Not fixed yet! @bcoe

bcoe:
@Stupidism I've tested @JaKXz's branch of your repo; if you switch the order of rewire and istanbul things should start working for you.

Stupidism:
Have you tried these three cmds?
```

yarn remove nyc
yarn clean
yarn add --dev nyc
```

[...]

PR #37, opened 2017-04-16
fix: normalize coverage data, to correct for bad coverage output

bcoe:
The reporting error discussed in #28 was caused by more counts existing for statements than mappings to code.

I'm not 100% sure of the root cause of this discrepancy, and might dig into things a little bit more before I land this.


bcoe:
the underlying issue was rewire being included in the wrong order; I've opened an issue to document this on the Istanbul website.

Issue #39, opened 2017-04-24
html reporter does not render source files


SamNelson:
On some of my files the html reporting is failing. Instead of rendering the source code annotated with coverage it is rendering the following stack trace:
```
Cannot read property 'start' of undefined
TypeError: Cannot read property 'start' of undefined
    at /Users/snels/projects/halo-api/node_modules/nyc/node_modules/istanbul-reports/lib/html/annotator.js:50:28
    at Array.forEach (native)
    at annotateStatements (/Users/snels/projects/halo-api/node_modules/nyc/node_modules/istanbul-reports/lib/html/annotator.js:46:33)
    at Object.annotateSourceCode (/Users/snels/projects/halo-api/node_modules/nyc/node_modules/istanbul-reports/lib/html/annotator.js:193:9)
    at HtmlReport.onDetail (/Users/snels/projects/halo-api/node_modules/nyc/node_modules/istanbul-reports/lib/html/index.js:217:39)
    at LcovReport.(anonymous function) [as onDetail] (/Users/snels/projects/halo-api/node_modules/nyc/node_modules/istanbul-reports/lib/lcov/index.js:24:24)
    at Visitor.(anonymous function) [as onDetail] (/Users/snels/projects/halo-api/node_modules/nyc/node_modules/istanbul-lib-report/lib/tree.js:34:30)
    at ReportNode.Node.visit (/Users/snels/projects/halo-api/node_modules/nyc/node_modules/istanbul-lib-report/lib/tree.js:123:17)
    at /Users/snels/projects/halo-api/node_modules/nyc/node_modules/istanbul-lib-report/lib/tree.js:116:23
    at Array.forEach (native)
```

Digging into the that stack trace it looks like html/annotator.js assumes that *fileCoverage.s* and *fileCoverage.statementMap* are always the same length.
```
var statementStats = fileCoverage.s,
        statementMeta = fileCoverage.statementMap;
    Object.keys(statementStats).forEach(function (stName) {
        var count = statementStats[stName],
            meta = statementMeta[stName],
            type = count > 0 ? 'yes' : 'no',
            startCol = meta.start.column,
            endCol = meta.end.column + 1,
            startLine = meta.start.line,
            endLine = meta.end.line,
            openSpan = lt + 'span class="' + (meta.skip ? 'cstat-skip' : 'cstat-no') + '"' + title('statement not covered') + gt,
            closeSpan = lt + '/span' + gt,
            text;
```

[...]

PR #40, opened 2017-04-24
added filter for statement statistics that contain null entries

SamNelson:
This should fix [issue #39](https://github.com/istanbuljs/istanbuljs/issues/39)

### DIFF
--- a/packages/istanbul-reports/lib/html/annotator.js
+++ b/packages/istanbul-reports/lib/html/annotator.js
@@ -41,9 +41,11 @@ function annotateLines(fileCoverage, structuredText) {
 }
 
 function annotateStatements(fileCoverage, structuredText) {
-    var statementStats = fileCoverage.s,
-        statementMeta = fileCoverage.statementMap;
-    Object.keys(statementStats).forEach(function (stName) {
+    var statementStats = Object.keys(fileCoverage.s).filter(function (stName) {
+      return fileCoverage.s[stName] !== null;
+    }),
+    statementMeta = fileCoverage.statementMap;
+    statementStats.forEach(function (stName) {
         var count = statementStats[stName],
             meta = statementMeta[stName],
             type = count > 0 ? 'yes' : 'no',
@@ -227,4 +229,3 @@ function annotateSourceCode(fileCoverage, sourceStore) {
 module.exports = {
     annotateSourceCode: annotateSourceCode
 };
-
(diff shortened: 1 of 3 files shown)


packages/istanbul-reports/lib/html/annotator.js
```diff
@@ -41,9 +41,11 @@ function annotateLines(fileCoverage, structuredText) {
 }
 
 function annotateStatements(fileCoverage, structuredText) {
-    var statementStats = fileCoverage.s,
-        statementMeta = fileCoverage.statementMap;
-    Object.keys(statementStats).forEach(function (stName) {
+    var statementStats = Object.keys(fileCoverage.s).filter(function (stName) {
+      return fileCoverage.s[stName] !== null;
```
JaKXz: would it be overly poor form to use `!= null` to capture any ones that are `undefined` as well? or would it be overkill to do `&& !== undefined`?

packages/istanbul-reports/lib/html/annotator.js
```diff
@@ -41,9 +41,11 @@ function annotateLines(fileCoverage, structuredText) {
...
+    }),
+    statementMeta = fileCoverage.statementMap;
```
JaKXz: I think this should be a `var statementMeta` line, it's kind of hard to grok imho.

bcoe:
@JaKXz I'm pretty sure that in normal circumstances, getting a `null` statement should be impossible;

@SamNelson are you able to share the codebase that you were bumping into issues on? I think if we switch the order of `istanbul` and `rewire` things might start working for you -- and we won't need this fix.

JaKXz:
@bcoe wouldn't it be good to have this passing anyways? in case there are other scenarios where there are null entries in statement statistics?

bcoe:
@JaKXz the problem is it would result in misreporting test coverage information -- except in bizarre situations, like the interaction we're seeing between `babel-plugin-rewire` and `babel-plugin-istanbul`, it should be impossible to have a `null` in a coverage map. If we'd been hiding this edge-case, we never would have found the bad interaction between these two libraries that was dropping coverage information on the ground.

bcoe:
@SamNelson please feel free to reopen this discussion if you're still bumping into issues 👍 we've rolled out a few changes to nyc/istanbul-lib-instrument that I think should solve your problem.

PR #46, opened 2017-04-29
chore: update travis config and lerna

JaKXz:


bcoe:
@JaKXz mind rebasing with `master` curious to see if we still see a drop in coverage.

JaKXz:
@bcoe curiously I am fully up to date.

PR #53, opened 2017-05-08
chore: update repo url

popomore:
All the urls should be refered to this repo that
can be easily found from npm.

Git url, bug url and homepage have been changed in package.json


bcoe:
@popomore thank you 👍

PR #60, opened 2017-05-19
fix: hoist statement counter for class variables

### LINKED ISSUE #59: Class variable instrumentation breaks name inferences
This function starts to return `false` after instrumentation:

```js
function test () {
  var foo = class {};
  var bar = class baz {};
  var qux = class { static name() {} };
  return foo.name === "foo" &&
         bar.name === "baz" &&
         typeof qux.name === "function";
}
```

This is the test for *variables (class)* in the [`node-compat-table` for function names](http://node.green/#ES2015-built-in-extensions-function--name--property). [See source](https://github.com/kangax/compat-table/blob/c0a3b882b27e4e9f4ef449a1c985ee7857326b94/data-es6.js#L13009:L13016).

The instrumentation yields:

```js
  var foo = (++cov_2mofekog2n.s[0], class {});
  var bar = (++cov_2mofekog2n.s[1], class baz {});
  var qux = (++cov_2mofekog2n.s[2], class {
    static name() {
      ++cov_2mofekog2n.f[1];
    }
  });
```

Do the `class` expressions need to be instrumented? For comparison, with `var foo = function() {};` the output is:

```js
  var foo = function () {
    ++cov_2mofekog2n.f[1];
  };
```

See also https://github.com/novemberborn/function-name-support which contains a regression test for `istanbul-lib-instruments` behavior.

novemberborn:
With `const foo = class {}`, `foo.name` should be `'foo'`. Without this commit the instrumentation causes `foo.name` to be an empty string.

Hoist the statement counter just like it's done for function and arrow expressions. Fixes #59.

### DIFF
--- a/packages/istanbul-lib-instrument/src/visitor.js
+++ b/packages/istanbul-lib-instrument/src/visitor.js
@@ -71,6 +71,12 @@ class VisitState {
         extractURL(node.trailingComments);
     }
 
+    // for these expressions the statement counter needs to be hoisted, so
+    // function name inference can be preserved
+    counterNeedsHoisting(path) {
+      return path.isFunctionExpression() || path.isArrowFunctionExpression() || path.isClassExpression();
+    }
+
     // all the generic stuff that needs to be done on enter for every node
     onEnter(path) {
         const n = path.node;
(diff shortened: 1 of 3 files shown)


bcoe:
@novemberborn great work 👍 thanks for diving into this.

bcoe:
@novemberborn this is now landed and available in `nyc@next` -- mind testing and making sure everything runs well for you?

novemberborn:
@bcoe going by https://github.com/novemberborn/function-name-support/issues/2 the failing AVA tests are now passing, which implies the instrumentation is no longer interfering. I haven't had a chance yet to actually update the test code though.

novemberborn:
Works great @bcoe.

PR #67, opened 2017-06-12
fix: set excludes to options.ignore on fileset

fengmk2:
It will make glob run faster.

---

- before:
  > fileset 7 files 7769ms
- after:
  > fileset 7 files 27ms


fengmk2:
Real project on https://github.com/eggjs/egg code coverage test.

## before

image

## after

image

fengmk2:
@bcoe Can you take a look for this change? It will help us a lot.

popomore:
@bcoe we can improve performance for coverage from this PR, can you review it?

JaKXz:
@fengmk2 I'm curious if there's a test case that should be added for this? Most likely not, but just in case; e.g. [some benchmarking](https://www.npmjs.com/package/benchmark) assertions would be a really cool addition so that we don't regress on this.

bcoe:
@fengmk2 awesome performance tweak 👍 sorry this took me so long to review.

PR #74, opened 2017-07-12
update to use correct node type

hzoo:
https://github.com/babel/babel/pull/5940 is a breaking change that fixes a bug, so making the change now since I hit it when trying to update in preset-env

https://github.com/babel/babel-preset-env/pull/372

### DIFF
--- a/packages/istanbul-lib-instrument/src/visitor.js
+++ b/packages/istanbul-lib-instrument/src/visitor.js
@@ -133,7 +133,7 @@ class VisitState {
                 ? (x) => T.memberExpression(x, T.numericLiteral(index), true)
                 : (x) => x
         );
-        return T.unaryExpression('++',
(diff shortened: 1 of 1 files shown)


bcoe:
@hzoo so I understand better, `unaryExpression` no longer includes `++` (which was a bug). `updateExpression` has always included `++` and is the correct expression type any ways?

This issue won't bite us until we upgrade to babel 7?

existentialism:
@bcoe yup!

hzoo:
Yep it was a bug (although breaking), and the reason I made the PR is because we are using Babel 7 in babel itself and we are using instanbul as well, and since this is a good change anyway it's fine to land for everyone without having to do a major bump.

bcoe:
@hzoo this is now released 👍 please let me know if it does the trick for you.

PR #80, opened 2017-08-23
fix: branch highlighting was going to end of file for HTML reports

vinteo:
With the logic in place, seems like `startLine` will always == `endLine`
so the condition will never be `false`. Coupled with the fact that
the original `endCol` from the `meta` object seems wrong (always less
than `startCol`), the hightlighting just went to the end of the file.

~~Fix this by removing the redundant logic and always setting the
`endCol` to the end of line using `text.originalLength()`~~

Fix this by removing the redundant logic and setting the
`endCol` to the end of line using `text.originalLength()` when
the given `endCol` is less than `startCol`

### DIFF
--- a/packages/istanbul-reports/lib/html/annotator.js
+++ b/packages/istanbul-reports/lib/html/annotator.js
@@ -57,13 +57,12 @@ function annotateStatements(fileCoverage, structuredText) {
 
         if (type === 'no' && structuredText[startLine]) {
             if (endLine !== startLine) {
-                endLine = startLine;
                 endCol = structuredText[startLine].text.originalLength();
             }
             text = structuredText[startLine].text;
             text.wrap(startCol,
                 openSpan,
-                startLine === endLine ? endCol : text.originalLength(),
+                startCol < endCol ? endCol : text.originalLength(),
                 closeSpan);
         }
     });
@@ -90,13 +89,12 @@ function annotateFunctions(fileCoverage, structuredText) {
 
         if (type === 'no' && structuredText[startLine]) {
             if (endLine !== startLine) {
-                endLine = startLine;
                 endCol = structuredText[startLine].text.originalLength();
             }
             text = structuredText[startLine].text;
             text.wrap(startCol,
                 openSpan,
-                startLine === endLine ? endCol : text.originalLength(),
+                startCol < endCol ? endCol : text.originalLength(),
                 closeSpan);
         }
     });
@@ -145,7 +143,6 @@ function annotateBranches(fileCoverage, structuredText) {
 
                 if (count === 0 && structuredText[startLine]) { //skip branches taken
                     if (endLine !== startLine) {
-                        endLine = startLine;
                         endCol = structuredText[startLine].text.originalLength();
                     }
                     text = structuredText[startLine].text;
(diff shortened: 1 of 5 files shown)


xiaofan2406:
@vinteo According to my test, same fix should be applied to `annotateStatements` as well.

bcoe:
@vinteo mind providing a screen-shot of the highlighting behavior you're seeing?

I wonder if the underlying problem could potentially be a source-map being remapped incorrectly, rather than a problem with the annotator.

before we make significant changes to this somewhat crufty logic, I'd be interested in getting a few tests in place, and I'd also like to make sure that the problem is in fact the reporting logic, and not a badly generated coverage file.

bcoe:
feel free to join the community slack I've setup, if you feel like helping coordinate work around this :) http://devtoolscommunity.herokuapp.com/

vinteo:
@bcoe Screenshot below shows the branch highlighting including lines that have been covered (as you can see by the numbers on the left margin)

image

The underlying problem could be the instrumenter/source maps as the `meta.end.column` for the line seems to be always less than `meta.start.column` (maybe specific to coffeescript), which doesn't make sense.

I used this https://github.com/gotwarlost/istanbul/pull/806 as a starting point, but I feel his fix is not right either as he is mixing columns with lines.

packages/istanbul-reports/lib/html/annotator.js
```diff
@@ -57,13 +57,12 @@ function annotateStatements(fileCoverage, structuredText) {
...
         if (type === 'no' && structuredText[startLine]) {
             if (endLine !== startLine) {
-                endLine = startLine;
                 endCol = structuredText[startLine].text.originalLength();
             }
             text = structuredText[startLine].text;
             text.wrap(startCol,
                 openSpan,
-                startLine === endLine ? endCol : text.originalLength(),
+                startCol < endCol ? endCol : text.originalLength(),
```
bcoe: I think this reads a bit better I like it 👍

vinteo: Yup, this means the existing logic is intact and only fixes the wrong scenario. Although still not ideal it means that if the source maps side of things gets fixed no change is required here.

vinteo:
@bcoe I added some tests, I hope we can get this in 😃

packages/istanbul-reports/test/html/annotator.js
```diff
@@ -32,5 +32,38 @@ describe('annotator', function () {
...
+          return '  var test = "test";';
+        }
+      });
+      annotated.annotatedCode[0].should
+        .equal('<span class="cstat-no" title="statement not covered" >  var test = "test";</span>');
+    });
+
+    // see: https://github.com/istanbuljs/istanbuljs/pull/80
+    it('handles function meta information with end column less than start column', function () {
+      var annotated = annotator.annotateSourceCode(getFixture('github-80b'), {
```
bcoe: this is great, I really appreciate the tests.

bcoe:
@vinteo I'm a little concerned about dropping:

`endLine = startLine;`

Mainly because I don't quite know what the original reasoning was behind this logic (it might be attempting to protect against a specific edge-case). Having said this, it seems like you've done a good job of familiarizing yourself with this part of the Istanbul codebase, and I really appreciate the tests.

I say we land this tomorrow, keep an eye out for any bug reports, and prepare to follow up with regression tests. Perhaps we can try to sync up in chat during the day tomorrow to coordinate the rollout?

vinteo:
@bcoe I think its low risk with removing that line as the `endLine` is not used at all further after the changes.  I think the initial intention was to keep the highlighting on the same line which is still maintained.

I'm based in Australia so it might be hard to coordinate a suitable time tomorrow (Saturday too for me) but I am willing to try and work around your schedule. Just ping me on slack.

Issue #85, opened 2017-09-14
no coverage report if source file is loaded by vm.runInContext


pangrr:
I think we need to add hook to vm.runInContext in a similar way that we hook vm.runInThisContext. Maybe for vm.runInNewContext as well.
Because some projects need to use vm.runInContext to create a testing context for code that is supposed to run in that context.

pangrr:
I am trying to make a PR to add hook to vm.runInContext but cannot find the global coverage variable to hook the vm context to. I guess the global coverage variable could be '__coverage__' but it is not initialized in when hook is called. Anyone give a hint?

pangrr:
It turns out ```` __coverage__ ```` is in the vm context. Thus it needs to be exported to global.

coreyfarrell:
It appears this was fixed by the linked PR's so I'm going to close this issue.  If further work is needed to resolve this issue let us know and we can reopen.

JoHuang:
I got the same issue when I use Jest. 
How to make hooks work in Jest?
Thanks

PR #90, opened 2017-09-22
hook vm.runInContext

pangrr:
fix issue [#85](https://github.com/istanbuljs/istanbuljs/issues/85) along with [this pull request](https://github.com/istanbuljs/nyc/pull/680) to nyc

I feel kind of weak on the coverage of this change. Please give suggestions and comments so I can improve this pull request. Thank you. 

Ran

### DIFF
--- a/packages/istanbul-api/lib/run-cover.js
+++ b/packages/istanbul-api/lib/run-cover.js
@@ -23,7 +23,12 @@ function getCoverFunctions(config, includes, callback) {
         reportingDir = path.resolve(config.reporting.dir()),
         reporter = new Reporter(config),
         excludes = config.instrumentation.excludes(true),
-        coverageVar = '__coverage__',
+        // The coverage variable below should have different value than
+        // that of the coverage variable actually used by the instrumenter (in this case: __coverage__).
+        // Otherwise if you run nyc to provide coverage on these files,
+        // both the actual instrumenter and this file will write to the global coverage variable,
+        // and provide unexpected coverage result.
+        coverageVar = '$$coverage$$',
         instOpts = config.instrumentation.getInstrumenterOpts(),
         sourceMapStore = libSourceMaps.createSourceMapStore({}),
         instrumenter,
@@ -88,7 +93,8 @@ function getCoverFunctions(config, includes, callback) {
     hookFn = function (matchFn) {
         var hookOpts = {
             verbose: config.verbose,
-            extensions: config.instrumentation.extensions()
+            extensions: config.instrumentation.extensions(),
+            coverageVariable: coverageVar
         };
 
         //initialize the global variable
--- a/packages/istanbul-api/lib/run-cover.js
+++ b/packages/istanbul-api/lib/run-cover.js
@@ -23,7 +23,7 @@ function getCoverFunctions(config, includes, callback) {
         reportingDir = path.resolve(config.reporting.dir()),
         reporter = new Reporter(config),
         excludes = config.instrumentation.excludes(true),
-        coverageVar = '$$cov_' + new Date().getTime() + '$$',
+        coverageVar = '__coverage__',
         instOpts = config.instrumentation.getInstrumenterOpts(),
         sourceMapStore = libSourceMaps.createSourceMapStore({}),
         instrumenter,
@@ -96,6 +96,9 @@ function getCoverFunctions(config, includes, callback) {
         reportInitFn();
 
         if (config.hooks.hookRunInContext()) {
+            hook.hookRunInContext(matchFn, transformer, hookOpts);
+        }
+        if (config.hooks.hookRunInThisContext()) {
             hook.hookRunInThisContext(matchFn, transformer, hookOpts);
         }
         disabler = hook.hookRequire(matchFn, requireTransformer, hookOpts);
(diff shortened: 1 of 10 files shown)


packages/istanbul-api/lib/run-cover.js
```diff
@@ -23,7 +23,7 @@ function getCoverFunctions(config, includes, callback) {
         reportingDir = path.resolve(config.reporting.dir()),
         reporter = new Reporter(config),
         excludes = config.instrumentation.excludes(true),
-        coverageVar = '$$cov_' + new Date().getTime() + '$$',
```
bcoe: I'm wondering if the reason why coverage dropped in CI might be that this variable changed? what's the motivation for this edit.

bcoe:
@pangrr this looks like a really solid start start to implementing this feature, thank you for the contribution. I would have to dig into why the coverage variable defaults to ` '$$cov_' + new Date().getTime() + '$$'`  in the Istnabul API ... but I'm wondering if changing this variable might be why coverage dropped significantly on this test run.

packages/istanbul-api/lib/run-cover.js
pangrr: @bcoe The reason for this change here is that:
1.  I add some unit tests for hookRunInContext under istanbul-api (because there are some similar unit tests here) and
2. hookRunInContext treats ```___coverage__```  (rather than ```$$cov__*```) as the coverage variable because ```___coverage__``` is the actual coverage variable used by istanbul, and
3. istanbul-api/run-cover.js is called in those unit test, so that ```$$cov__*``` becomes the coverage variable in those tests.
4. In order to make the unit tests pass, I changed the coverageVar here.

pangrr:
@bcoe Thank you for the feedback. Yeah, I also wonder why ```$$cov_*``` is inherited from the old istanbul where new istanbul is using ```__coverage___```. It seems that ```$$cov_*``` is only used under istanbul-api which is not used by nyc.

bcoe:
@pangrr get a chance to dig into this any more? mainly curious about the drop in coverage.

pangrr:
@bcoe Apologize for the very late reply. I was distracted from other works. 

- As you said, the significant coverage drop is triggered by the change of coverageVar in run-cover.js. I created [another PR](https://github.com/istanbuljs/istanbuljs/pull/98) with basically the same feature as this PR but without change of coverageVar (thus we lose a little unit test coverage).

- As to this PR, a lot of files lose total coverage. Most of those files don't even use run-cover.js. From the CI log, I see the unit tests get executed for those files.

- Based on the observation above, I wonder:

1. How the coverage is generated in the CI?
2. How istanbul-api and other packages are related? Because for example, many files under istanbul-lib-reports lose total coverage, while istanbul-lib-reports doesn't depend on istanbul-api.
2. What is the purpose of istanbul-api package? Because this package is not used by any other package in either istanbuljs nor nyc.

Could you please provide some hint?

pangrr:
Based on pull request #101 and #100 it seems that ```$$cov*$$``` is the magic keywords. But why?

pangrr:
According to [thisCI job report](https://travis-ci.org/istanbuljs/istanbuljs/jobs/282032476) as compared to [another report with normal coverage](https://travis-ci.org/istanbuljs/istanbuljs/jobs/281382360):

- Coverage drops consist of:
    - loss of coverage for the whole istanbul-api package and
    - loss of coverage for some of istanbul-reports files
- The reason of loss of coverage for istanbul-api may be because both nyc (which provides the actual coverage) and istanbul-api/run-cover.js (which is used to run every unit test for files in istanbul-api) use ```global.__coverage__``` and run-cover.js messed up the coverage data.
- Why this affects istanbul-reports? Is it because istanbul-reports tries to access something related to istanbul-api's coverage data?

bcoe:
@pangrr sorry for the slow reply; sneaking in OSS between other life stuff lately.

Here's a thought, why don't you make it so you pass the coverage variable name as an option into `hookRunInContext` that way you can keep the alternate coverage name in tests, which should bump up coverage. Then, for nyc's case, just default it to `__coverage__` if no alternate value is provided.

pangrr:
@bcoe Thank you for this suggestion. I thought about that before, my hesitation was that it feels like an over kill to add a parameter just for testing.
On the other hand, the global coverage variable is configurable too. Thus I reckon it may be a good idea to make the coverage variable configurable for the hook.
Will do the change in this pull request.

bcoe:
@pangrr this is looking pretty solid now 😄 no drop in coverage it would appear? and coverage for the new feature as expected? When you have a bit of time, perhaps join the community slack:

http://devtoolscommunity.herokuapp.com/

and help prod me into releasing this? I'm currently working through a backlog of issues (they collect up pretty fast) and have a bit of work I'm wrapping up on my other library yargs _but_ I'm excited to start landing some of your work.

pangrr:
@bcoe Thank you for inviting me to the community.
The coverage for files executed in vm.runInContext is as expected.

How can I help to release this?

bcoe:
@pangrr this is amazing work, sorry it took me so long to land; I'll work on getting a new version of the pertinent libraries released soon ... I'm hoping to patch a few more bugs before the end of the weekend.

bcoe:
@pangrr your work is out in `+ istanbul-lib-hook@1.1.0`

PR #93, opened 2017-09-27
Fix hook runInThisContext

### LINKED ISSUE #89: hookRunInThisContext doesn't use current runInThisContext API
#The `vm.runInThisContext` hook assumes the second argument to `runInThisContext` [is a string](https://github.com/istanbuljs/istanbuljs/blob/master/packages/istanbul-lib-hook/lib/hook.js#L139). However, the argument has been an options object for quite a while now (I believe since Node 0.10 or 0.12). This prevents Istanbul from instrumenting modules instantiated by calling `Module#_compile` (or by calling `runInThisContext` using the proper API).

CodeTroopers:
Fix #89 hook runInThisContext to use options as argument instead of  filename (cf [vm documentation](https://nodejs.org/api/vm.html#vm_vm_runinthiscontext_code_options))

Delete unusable and unused hook on createScript : instead use constructor [new Script](https://nodejs.org/api/vm.html#vm_new_vm_script_code_options)

### DIFF
--- a/packages/istanbul-api/lib/run-cover.js
+++ b/packages/istanbul-api/lib/run-cover.js
@@ -3,198 +3,197 @@
  Copyrights licensed under the New BSD License. See the accompanying LICENSE file for terms.
  */
 var path = require('path'),
-    fs = require('fs'),
-    mkdirp = require('mkdirp'),
-    matcherFor = require('./file-matcher').matcherFor,
-    libInstrument = require('istanbul-lib-instrument'),
-    libCoverage = require('istanbul-lib-coverage'),
-    libSourceMaps = require('istanbul-lib-source-maps'),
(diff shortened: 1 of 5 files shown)


CodeTroopers:
Strange error on 4 unit tests : Uncaught TypeError: Cannot read property 'f' of undefined
Help would be appreciated

bcoe:
@CodeTroopers thanks for digging into this; If you don't mind, let's split this into two pull requests.

One that corrects the `runInThisContext` logic, and one that removes `hookCreateScript` ... I'm not quite following why you removed this API method though? Looking at tests it seems as though the feature was behaving as expected?

CodeTroopers:
Close for a more clean pull request

PR #95, opened 2017-09-28
deps: upgrade debug package

evan-scott-zocdoc:
Due to https://nodesecurity.io/advisories/534

It's a major version bump, but looking at the changelog I don't
see any breaking changes of note.


bcoe:
@evan-scott-zocdoc 👍 thanks for the contribution.

PR #98, opened 2017-09-29
another pull request to add hook to vm.runInContext

pangrr:
This PR is created to deal with the coverage loss issue of [another PR](https://github.com/istanbuljs/istanbuljs/pull/90) of the same purpose.
The difference between this PR and the other PR is that, this PR keep the value of coverageVar in istanbul-api/lib/run-cover.js and remove a little bit of unit test on hookRunInContext.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,5 @@ node_js:
   - '6'
   - 'stable'
 
+before_install: npm i yarn -g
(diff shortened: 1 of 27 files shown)


bcoe:
@pangrr whoops; I commented on the other pull, which I believe this would override. Same comment applies, perhaps we could make the `__coverage__` variable configurable, such that your tests are covered as expected?

PR #108, opened 2017-10-16
feat: add support for ignoring entire files

kentcdodds:
Sometimes in my projects I'll have one or two files that I don't want to have coverage for. It's annoying to have to edit the config to just ignore one or two files (especially if that config is hidden away in another tool).

This PR adds a feature for ignoring entire files by allowing you to add a comment: `/* istanbul ignore file */` and it will skip instrumenting the entire file.

To do this, I create another regex especially for this full file ignore, then in the program enter and exit visitor, we check whether the parent of the program node (the file node) has any comments which match this regex. If it does, then we'll skip doing any instrumentation.

This PR also tests this functionality with a new `specs/ignore.yaml`. It requires adding a new option called `noCoverage` which functions similar to `generateOnly` except it adds an assertion that the input code is the same as the output (because it hasn't been altered).

I've tested this locally and it works great!
 
screen shot 2017-10-16 at 5 26 30 pm

I normally would have opened an issue for this, but I thought it'd be pretty small :) I'm happy to iterate on this a bit.

### DIFF
--- a/packages/istanbul-lib-instrument/src/visitor.js
+++ b/packages/istanbul-lib-instrument/src/visitor.js
@@ -5,6 +5,8 @@ import template from 'babel-template';
 
 // istanbul ignore comment pattern
 const COMMENT_RE = /^\s*istanbul\s+ignore\s+(if|else|next)(?=\W|$)/;
+// istanbul ignore file pattern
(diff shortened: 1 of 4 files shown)


bcoe:
@kentcdodds your work is released in `istanbul-lib-instrument: 1.8.0 => 1.9.0`.

PR #115, opened 2017-11-14
Build/fix travis appveyor

JaKXz:


bcoe:
@JaKXz want to take another crack at kicking this off, I think the early `9.x` release might have had some problems.

JaKXz:
@bcoe unfortunately looks like we're still waiting on that fix. No idea why node v4 on windows isn't co-operating though :(

bcoe:
@bcoe I believe the fix is supposed to be out this week to npm 👍

JaKXz:
Yay Travis is passing now! Thank you npm folks!

bcoe:
@JaKXz looking good 👍 mind kicking off another appveyor build when you have two seconds; I wonder why the Node 4 build failed.

JaKXz:
@bcoe no dice :(

CodeTroopers:
The error is referenced [here](https://github.com/lerna/lerna/issues/999)

bcoe:
@CodeTroopers, @JaKXz sounds like there's potentially a workaround for this issue in the thread that @CodeTroopers linked to? would love an update to this patch if anyone has a moment.

CodeTroopers:
@bcoe I'm not sure of the need to pass time for this issue. It concern only NodeJS4 and [the end support is in April](https://github.com/nodejs/Release)

PR #122, opened 2017-12-20
Fixes proper passing of the preserveComments option to babel

karottenreibe:
I noticed that passing preserveComments: false to Istanbul doesn't do anything. Comments were never removed. I fixed this. The option was simply never used anywhere. Also wrote a test for it and fixed one other test

Hope my changes are OK, otherwise please let me know!


bcoe:
@karottenreibe great catch. I apologize for the slow turn around, I've been dealing with personal issues for the first part of this year.

PR #128, opened 2017-12-29
Fix fileCoverage of undefined failure

kylekatarnls:
Avoid output to be undefined and throw `Cannot read property 'fileCoverage' of undefined`


kylekatarnls:
@bcoe @JaKXz Hi maintainers, any plan to merge this fix? Do you need something more?

bcoe:
@kylekatarnls I apologize for the slow turnaround on this issue, I've been dealing with a family tragedy.

Can you provide an example that would result in `ee.exit(path);` returning `undefined` -- I wonder if this fix might just be masking an issue.

bcoe:
@kylekatarnls friendly reminder, any chance you could provide a reproduction for this behavior?

PR #131, opened 2018-01-05
fix: fix instrumentation when a function is called Function

addaleax:
Hi, it’s been a while! 👋 Hope y’all are doing great!

This patch makes the code a little more robust by not assuming that the instrumented code contains no function named `Function`.


bcoe:
@addaleax thank you for the contribution 👍 I'll work on having something out soon.

bcoe:
@addaleax this is released now, please let me know if this works for you.

addaleax:
@bcoe Works like a charm :smile_cat:

Issue #92, opened 2017-09-26
Coverage init no longer hoisted to the top in Babel 7.0.0-beta


loganfsmyth:
Heya, just came across this so I figured I'd pass it along. 

https://github.com/istanbuljs/istanbuljs/blob/2ce8974/packages/istanbul-lib-instrument/src/visitor.js#L519

is now breaking on Babel 7.x beta because we added a new blockHoist level `4` to work around some issues.

You'll need to bump that up for things to keep working.

That said, we've been really hoping that in the long run we can remove the concept of blockHoist. I don't know if it'll happen, but we'll see.

Have you all considered approaches that avoid the need for it? For instance if you use hoisting to your advantage and made that a function declaration and called it every time you needed the object, you could lazy-init the coverage object on first access and avoid the need got blockHoist.

bcoe:
@loganfsmyth thanks for the heads up; in the short term we'll probably just bump up to `4` when we upgrade to Babel 7.

Would definitely entertain eliminating the need for `blockHoist` in the future -- but would definitely want to do some performance benchmarking to see if there's any noticeable slowdown to instrumentation or execution as a result of refactoring ... realistically, I also don't have too many cycles in the immediate future to take on this refactor unless someone else dives into it.

loganfsmyth:
The blockHoist value has been increased in https://github.com/istanbuljs/istanbuljs/pull/135 so this issue could potentially be closed.

That said, it might be good to have this or another issue to track removing `blockHoist` in general. One downside of the current approach is that it's only 100% effective if Babel is the only tooling performing transformations. Now that Webpack has good support for ES6 module syntax for instance, fewer people are using Babel's compiled modules, which means that it isn't possible for `_blockHoist` in ensure that the `cov` structure is initialized before imports are executed. The same is true of native module support landing in browsers.

This means that it is possible for dependency cycles between files to run functions before the file coverage structure has been initialized.

My suggestion would still be to use a lazy-initialized function declaration to handle hoisting, e.g.

```
var COVERAGE_VAR = (function () {
    var path = PATH,
        hash = HASH,
        global = (new Function('return this'))(),
        gcv = GLOBAL_COVERAGE_VAR,
        coverageData = INITIAL,
        coverage = global[gcv] || (global[gcv] = {});
    if (coverage[path] && coverage[path].hash === hash) {
        return coverage[path];
    }
    coverageData.hash = hash;
    return coverage[path] = coverageData;
})();
```
could instead be
```
[...]

PR #135, opened 2018-01-12
Fix compatibility with babel 7

danez:
Babel 7 does not return program scope from getFunctionParent().

This fix also still works with babel 6

https://github.com/babel/babel/pull/5923

### DIFF
--- a/packages/istanbul-lib-instrument/src/read-coverage.js
+++ b/packages/istanbul-lib-instrument/src/read-coverage.js
@@ -29,7 +29,7 @@ export default function readInitialCoverage (code) {
                 if (!magicValue.confident || magicValue.value !== MAGIC_VALUE) {
                     return;
                 }
-                covScope = path.scope.getFunctionParent();
+                covScope = path.scope.getFunctionParent() || path.scope.getProgramParent();
                 path.stop();
             }
         }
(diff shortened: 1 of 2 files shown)


bcoe:
@danez thank you for pitching in on this \o/ one other outstanding issue we might want to correct while we're at it:

https://github.com/istanbuljs/istanbuljs/issues/92

@loganfsmyth anything else that you think we should have on our mind for this upgrade?

loganfsmyth:
I'd say it should be fine to just increment the blockHoist value to 5 for 7.x, but in the long run I'd still love for this plugin to change to function calls to avoid using it entirely, assuming that isn't a massive performance hit.

bcoe:
@danez awesome, I will do my best to land this sometime this week 👍

bcoe:
@danez I apologize for taking so long to land this -- I've been dealing with a personal tragedy over the past few months, and am just starting to do some open-source again.

I will work on having a release out the door soon for this.

loganfsmyth:
Nothing to apologize for. Thanks for taking the time to merge :)

bcoe:
@danez this is released now, please take it for a spin and let me know if you run into any issues.

PR #138, opened 2018-01-31
Changed column header from "Uncovered Lines" to "Uncovered Line #s"

shazron:
To improve the understanding of this column.

For example, you might have a value "139,140" which you could misunderstand as your file having 139,140 lines as uncovered (however unlikely).

See example https://github.com/istanbuljs/nyc/issues/35
Inspired by: https://github.com/gotwarlost/istanbul/pull/829


shazron:
AppVeyor has an an unrelated error which I noticed that all new PRs also suffer from:
"Error: Cannot find module 'highlight.js'"

Notice that there is no appveyor badge on the main README that would have highlighted this problem in master as well. Might be related -> https://github.com/istanbuljs/istanbuljs/pull/115

bcoe:
@shazron this is released now, please take it for a spin and let me know if it does the trick for you.

PR #156, opened 2018-03-31
Revert "fix: pass correct args to tranformer (#153)"

bcoe:
Reverts istanbuljs/istanbuljs#154 @kevinkir, @CodeTroopers reading back through #99 I note that the signature of transformer would need to change in a bunch of other places -- so landing #154 creates an inconsistent API ... I think at this point we pretty much need to just update the documentation to reflect the API change, we could release this as a breaking change and and revert `latest` back to a slightly older version of the `1.x` version of the library.

@kevinkir want to take a stab at updating the docs and API?


kevinkir:
@bcoe sounds good. I'll try to get to it within the next few days.

PR #168, opened 2018-05-18
fix: ensure using correct context

### LINKED ISSUE #167: [istanbul-reports] html report crash with babel 7
Using `babel` 7 and `babel-plugin-istanbul` I get an crash on [this](https://github.com/istanbuljs/istanbuljs/blob/master/packages/istanbul-reports/lib/html/index.js#L174)

image

Changed it to use the `that` variable and it works.

stoffeastrom:
Fixes #167


bcoe:
this definitely seems like the correct behavior to me 👍 thanks for the contribution.

Issue #163, opened 2018-05-02
`process.cwd` is a function, not a string


Axure:
https://github.com/istanbuljs/istanbuljs/blob/ac824a46b25a815977394d96843a31657a2be77b/packages/istanbul-lib-source-maps/lib/pathutils.js#L12

This would possibly cause errors, like:

```
Path must be a string. Received [Function]
TypeError: Path must be a string. Received [Function]
    at assertPath (path.js:28:11)
    at Object.resolve (path.js:1171:7)
```

PR #171, opened 2018-05-21
fix: process.cwd is a function not a string (#163).

coreyfarrell:


bcoe:
@coreyfarrell thanks!

PR #174, opened 2018-05-30
chore: upgrade babel in instrumenter

bcoe:


bcoe:
CC: @hzoo, @loganfsmyth, how is this upgrade looking? I want to be fairly careful not to break the world.

packages/istanbul-lib-instrument/package.json
```diff
@@ -14,11 +14,11 @@
     "prepublish": "npm run release"
   },
   "dependencies": {
-    "babel-generator": "^6.18.0",
-    "babel-template": "^6.16.0",
-    "babel-traverse": "^6.18.0",
+    "@babel/generator": "^7.0.0-beta.49",
```
loganfsmyth: I wouldn't use `^` in any of these versions since you never know if we'll make any final breaking changes that may or may not affect this.

packages/istanbul-lib-instrument/src/read-coverage.js
```diff
@@ -14,7 +14,10 @@ export default function readInitialCoverage (code) {
         allowReturnOutsideFunction: true,
         allowSuperOutsideMethod: true,
         sourceType: "script", // I think ?
-        plugins: ["*"]
+        plugins: [
+          "*",
```
loganfsmyth: The parser doesn't have `*` anymore. Some parser plugins conflict with others, so we require an explicit list of parser plugins to enable.

bcoe:
```/home/travis/build/istanbuljs/istanbuljs/node_modules/@babel/traverse/lib/path/index.js:109
  static get({
             ^
SyntaxError: Unexpected token {
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Module.replacementCompile (/home/travis/build/istanbuljs/istanbuljs/node_modules/nyc/node_modules/append-transform/index.js:58:13)```

File: packages/istanbul-lib-instrument/package.json [hunk 3]
```diff
@@ -14,11 +14,11 @@
...
+    "@babel/template": "^7.0.0-beta.49",
+    "@babel/traverse": "^7.0.0-beta.49",
     "babel-types": "^6.18.0",
```
loganfsmyth: Did this get missed?

bcoe:
@loganfsmyth it looks like Node 4 support is dropped in Babel@7?

loganfsmyth:
@bcoe Yes it is, so that's a good point. We've doing >= 6.9 for Babel 7.

packages/istanbul-lib-instrument/package.json
```diff
@@ -14,11 +14,11 @@
...
-    "babel-generator": "^6.18.0",
-    "babel-template": "^6.16.0",
-    "babel-traverse": "^6.18.0",
-    "babel-types": "^6.18.0",
-    "babylon": "^6.18.0",
+    "@babel/generator": "7.0.0-beta.49",
+    "@babel/template": "7.0.0-beta.49",
+    "@babel/traverse": "7.0.0-beta.49",
+    "@babel/types": "7.0.0-beta.49",
+    "babylon": "7.0.0-beta.47",
```
loganfsmyth: Oh we renamed, which is why the versions don't match. It's just `@babel/parser` now.

bcoe: interesting, switching to `@babel/parser` seems to break shebang support:

```bash
  1) statement.yaml/shebang code
       covers line and one branch:
     AssertionError: Error compiling
1: #!/usr/bin/env node
2: 
3: var x = args[0] > 5 ? args[0] : "undef";
4: output = x;
5: 
Invalid or unexpected token: expected false to be truthy
```

is there an additional plugin that's required? seems to be  a regression from babylon@7.0.0-beta.47.

loganfsmyth: Hmm interesting. I'll take a look. I did land a change for shebang handling, but I can't tell from that error if it's a bug, or just not expecting the new logic.

bcoe: interesting, it doesn't seem to be shebangs in general, this works:

```js
// #!/usr/bin/env node
var x = args[0] > 5 ? args[0] : "undef";
output = x;
```

but this doesn't work:

```js
#!/usr/bin/env node
var x = args[0] > 5 ? args[0] : "undef";
output = x;
```

loganfsmyth: Okay, I see what's up. In Babel 6 the interpreter directive was basically silently converted into a LineComment in the AST. 

In 7.x, we changed is that now if your input file had a shebang, it actually gets passed to the code generator and generated as a shebang again. That's causing problems for your logic here: https://github.com/istanbuljs/istanbuljs/blob/ac824a46b25a815977394d96843a31657a2be77b/packages/istanbul-lib-instrument/test/util/verifier.js#L127 that means the interpreter directive is no longer at the start of the file, and thus a syntax error.

loganfsmyth: Maybe your tests should just strip directives in that logic?

hulkish:
@bcoe this should have been a major version bump. Tbis breaks babel 6 users. Please see https://github.com/istanbuljs/istanbuljs/issues/206

Issue #176, opened 2018-05-31
istanbul-lib-report throws on path names containing Object.prototype members


msluther:
If you have a setup like this:
/lib/constructor/index.js

Then when `istanbul-lib-report` comes along, the summarizer will run this line of code:
https://github.com/istanbuljs/istanbuljs/blob/master/packages/istanbul-lib-report/lib/summarizer.js#L174
basically asking for `({}).constructor`, which is defined as its in the Object.prototype.  it should have fallen into the `if (!parent) {` statement to properly create the ReportNode, but doesn't.  It falls through and tries to run `parent.addChild(...)` and throws.
`TypeError: parent.addChild is not a function`

The work around is to change my file system layout to rename my `constructor` directory to something else.  So I'm not blocked per-se because of this, but its making me bend a little funny to do things because of a code coverage reporting tool.

PR #177, opened 2018-06-01
fix: use null prototype for map objects

bmeck:
should fix https://github.com/istanbuljs/istanbuljs/issues/176


msluther:
Verified that this fixes my use case.

bcoe:
@bmeck to get your tests to pass, I believe you also need to update a couple `{}` assignments in `coverage-map.js`.

bmeck:
@bcoe tests are passing locally, but yes those probably need this as well.

PR #180, opened 2018-06-04
Update dependencies

coreyfarrell:
This is probably easiest to review the changes one commit at a time.  The first commit updates production dependencies where the latest version is not allowed by package.json.  The second commit does the same for development dependencies.  The third commit performs all remaining updates.  This helps lerna hoist more packages since it won't always packages that have compatible but different version specifiers, like `^2.8.0` vs `^2.9.5`.

The only module not updated to latest is `source-map`.  It is updated to the highest version which supports node.js 6, the latest version requires node.js 8.  One change that deserves some scrutiny is the replacement of micromatch with minimatch.  This was done since `nyc` already (indirectly) uses minimatch to perform glob matching.

This combined with changes for nyc reduce the total size of `npm i -D nyc` by 10% - 21MB instead of 23MB installed by `nyc@^12.0.2`.  This updated code was run against local tests and `nyc` tests, all pass.


stoffeastrom:
@coreyfarrell @bcoe 

I think this update broke the `shouldInstrument` logic. Upgrading `nyc` to `13.0.0` and nothing gets included in the coverage. Reverting to `12.0.2` and it works. I guess the switch to `minimatch` and the code changes breaks things. I will do some more investigation

stoffeastrom:
I think I found the issue. I was generating a pattern

```javascript
*(a/a|b/b|c/c)/src/**/*.js
```
which was working before but not now.

stoffeastrom:
Changing to 

```javascript
*(a|b|c)/*/src/**/*.js
```
should solve it

coreyfarrell:
Maybe I'm misunderstanding but I just created a file `some-a/a/src/sub/file.js`.  In my `package.json` I included:
```json
{
  "nyc": {
    "include": ["*(a/a|b/b|c/c)/src/**/*.js"]
  }
}
```

This did not generate coverage for `some-a/a/src/sub/file.js` under nyc 12 or 13.  Maybe I'm misunderstanding what filename should be covered by this statement?  I was able to get coverage for the file using the include pattern `*{a/a,b/b,c/c}/src/**/*.js` (this worked in nyc 12 and 13).

stoffeastrom:
Weird, It was a big pattern generated for lerna packages so I might have missed something else. I got it working so didn't dive into it more. Thanks for your reply @coreyfarrell

coreyfarrell:
@stoffeastrom if you have a chance and figure out a way to reproduce the issue you had with include patterns this would be helpful.  I'd like to try finding out what the change is and at minimum document it.

PR #182, opened 2018-06-08
fix: Instrument ObjectMethod's.

coreyfarrell:
nyc currently reports the following code as containing 0 functions.  Add
ObjectMethod visitor to istanbul-lib-instrument so these are seen.

```js
const obj = {
	get prop() { return 0; }
	fn() {}
	*gen() {}
};
```


bcoe:
@coreyfarrell thank you for figuring this out! I believe there's an open issue (perhaps on nyc) related to this?

PR #184, opened 2018-06-18
test-exclude: bump read-pkg-up dependency

LinusU:
The only breaking change is that we dropped support for Node.js 4, which shouldn't affect this module since it seems like it only supports Node.js 6 and up.

### DIFF
--- a/packages/test-exclude/package.json
+++ b/packages/test-exclude/package.json
@@ -35,7 +35,10 @@
   "dependencies": {
     "arrify": "^1.0.1",
     "minimatch": "^3.0.4",
-    "read-pkg-up": "^3.0.0",
+    "read-pkg-up": "^4.0.0",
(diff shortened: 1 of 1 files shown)


coreyfarrell:
Actually test-exclude still works with node.js 4.  Although NYC and other parts of istanbuljs require node.js 6 this doesn't yet.  nyc is not the only [dependent](https://www.npmjs.com/browse/depended/test-exclude) of this module.

I think we should probably mark this as a breaking change in the commit message so it triggers a semver-major for test-exclude.

LinusU:
Sounds good! I'll also add `engines` to the package.json to explicitly state what version of Node.js is supported, sounds good?

LinusU:
Fixed 👌

coreyfarrell:
Thanks for the update @LinusU.

PR #190, opened 2018-06-27
feature: Update for babel 7.

coreyfarrell:
BREAKING CHANGE: This updated configuration requires babel-plugin-istanbul 5.0.0.

Now requires node >=6 per dependencies.

### DIFF
--- a/packages/nyc-config-babel/README.md
+++ b/packages/nyc-config-babel/README.md
@@ -14,7 +14,7 @@ Then write a `.babelrc` that looks something like this:
 
 ```json
 {
-  "presets": ["env", "..., etc."],
+  "presets": ["@babel/env", "..., etc."],
   "plugins": ["istanbul"]
 }
 ```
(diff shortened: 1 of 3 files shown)


packages/nyc-config-babel/package.json
```diff
@@ -26,8 +26,11 @@
   "publishConfig": {
     "access": "public"
   },
+  "engines": {
+    "node": ">=6"
+  },
   "peerDependencies": {
-    "babel-plugin-istanbul": "*",
+    "babel-plugin-istanbul": ">=5",
     "babel-register": "*"
```
JaKXz: I think this needs to be changed as well

packages/nyc-config-babel/README.md
```diff
@@ -14,7 +14,7 @@ Then write a `.babelrc` that looks something like this:
 
 ```json
 {
-  "presets": ["env", "..., etc."],
+  "presets": ["@babel/preset-env", "..., etc."],
```
JaKXz: NITPICK: I think `@babel/env` is a valid and shorter alias.

coreyfarrell: Appears so I'm not sure why I previously remember having trouble with this.

Updated for this and your other finding, thanks for the review!

Issue #191, opened 2018-06-27
Coverage Broken: parent.addChild is not a function


fabriciojso:
Run command: `jest --coverage`
Failed to write coverage reports:
        ERROR: TypeError: parent.addChild is not a function
        STACK: TypeError: parent.addChild is not a function
    at /Users/fabriciojso/github/tcc/node_modules/istanbul-lib-report/lib/summarizer.js:181:16
    at Array.forEach (<anonymous>)
    at toDirParents (/Users/fabriciojso/github/tcc/node_modules/istanbul-lib-report/lib/summarizer.js:171:10)
    at Reporter.createPackageSummary [as summarizer] (/Users/fabriciojso/github/tcc/node_modules/istanbul-lib-report/lib/summarizer.js:241:22)
    at Reporter.write (/Users/fabriciojso/github/tcc/node_modules/istanbul-api/lib/reporter.js:81:21)
    at /Users/fabriciojso/github/tcc/node_modules/jest-cli/build/reporters/coverage_reporter.js:181:18
    at Generator.next (<anonymous>)
    at step (/Users/fabriciojso/github/tcc/node_modules/jest-cli/build/reporters/coverage_reporter.js:75:30)
    at /Users/fabriciojso/github/tcc/node_modules/jest-cli/build/reporters/coverage_reporter.js:86:15
    at <anonymous>

rodrigooler:
PR was created with the above problem done by @fabriciojso  https://github.com/istanbuljs/istanbuljs/pull/193

coreyfarrell:
I'm not able to reproduce the error with the linked repository.  I did the following using node.js v10.15.0:
```sh
git clone https://github.com/fabriciojso/monetify-imagini
cd monetify-imagini
npm i
npm t
```
The test `src/core/interface/Application.test.ts` failed but the istanbul reporting seemed to work:
```
--------------------|----------|----------|----------|----------|-------------------|
File                |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
--------------------|----------|----------|----------|----------|-------------------|
All files           |    83.58 |    76.47 |    79.17 |    84.85 |                   |
 core/component     |      100 |      100 |      100 |      100 |                   |
  GroupComponent.ts |      100 |      100 |      100 |      100 |                   |
 core/constructor   |      100 |      100 |      100 |      100 |                   |
  Background.ts     |      100 |      100 |      100 |      100 |                   |
[...]

PR #193, opened 2018-06-27
Proposal for resolution of the bug reported (#191)

fabriciojso:
An error was encountered in issue #191, and this implementation proposes a solution to this error.


bcoe:
@fabriciojso thanks for digging into this ... what's weird to me, what's `parent` equal to when the method `addChild` does not exist? It seems odd that parent would be truthy, but not a `ReportNode` -- if you don't mind adding some logging, and digging into what the but might be, it feels like there's a chance we're masking a deeper issue.

coreyfarrell:
@fabriciojso are you able to address the comment by @bcoe?  I'm concerned that your patch could cause a bug to be suppressed rather than fixed.

PR #198, opened 2018-07-04
chore: Specify node >= 6 in all package.json files.

coreyfarrell:
This updates packages which have already gotten a semver-major bump since `latest` npm tag.


coreyfarrell:
@JaKXz why do you say that?  Not really against switching it but honestly I feel like this is a housekeeping commit or maybe `docs`.

JaKXz:
@coreyfarrell because I think it's something that should trigger at least a patch on the packages through standard-version.

And technically this might break builds for users so maybe it should be major versions everywhere?

coreyfarrell:
@JaKXz So for example `istanbul-api` is published to npm, latest = 1.3.1, next = 2.0.3.  I had thought that this should trigger 2.0.4 since 2.0.0 - 2.0.3 are 'prerelease only', so we don't want to skip to 3.0.0.

Unless I'm misunderstanding what the commit tags would do.  If I mark everything breaking would the release tools understand that `istanbul-api@2.x.x` has never been released so don't skip to 3.0.0?

JaKXz:
Ah - I didn't know the current state of the pre-release versions. This makes sense to me then, but, to trigger the patch update I think you need to use `fix` in the commit message @coreyfarrell.

bcoe:
@JaKXz any prefix will trigger a patch update, the prefix `fix` just ensures that it's documented in the CHANGELOG.

PR #200, opened 2018-07-10
feat: New option coverageGlobalScope.

### LINKED ISSUE #199: Proposal: option to put __coverage__ on window.top (for iframes)
Hello! 👋 

### The Problem
I'm using `nyc` to gather code coverage for a web page which makes heavy use of iframes. In addition, it also makes heavy use of ShadowDOM. The end result is that the page has an unknown number of iframes in an unknown number of shadow roots. Crawling the entire page to discover each iframe is possible, but tedious and slow.

This is a problem, because it makes harvesting the `__coverage__` object from each iframe very difficult.

### The Solution
I would like to propose a way to tell `istanbul-lib-instrument` to always put coverage on [`window.top`](https://developer.mozilla.org/en-US/docs/Web/API/Window/top). This would ensure that I can always grab coverage from all of my iframes in one place, regardless of how many iframes there are and even if there are nested iframes.

It looks like [istanbul-lib-instrument/src/visitor.js#L485](https://github.com/istanbuljs/istanbuljs/blob/71b815d111af5181196173f8af94f14510bb5f7b/packages/istanbul-lib-instrument/src/visitor.js#L485) is where the current parent of the `__coverage__` object is defined, and would be where I'd need to implement this option.

I don't yet have a plan of attack, because I wanted to first see if such an option would be welcome as a Pull Request.

### Questions
1. Is my use case reasonable? Should `istanbul-lib-instrument` even try to support it?
2. Is my proposed solution reasonable?
3. Is there an alternative or simpler solution that I am unaware of?

Thanks for your time,
Alex

coreyfarrell:
This new option allows modification of the function which finds the
global variable scope.  The default global scope is `this`, resulting in
`(new Function('return this))()` to retrieve the global scope.  This can
now be changed, for example `window.top` could be used to always use the
top level IFRAME in a web browser.

Fixes #199

### DIFF
--- a/packages/istanbul-lib-instrument/api.md
+++ b/packages/istanbul-lib-instrument/api.md
@@ -106,6 +106,8 @@ The exit function returns an object that currently has the following keys:
 -   `sourceFilePath` **[string][9]** the path to source file (optional, default `'unknown.js'`)
 -   `opts` **[Object][8]** additional options (optional, default `defaultProgramVisitorOpts`)
     -   `opts.coverageVariable` **[string][9]** the global coverage variable name. (optional, default `__coverage__`)
+    -   `opts.coverageGlobalScope` **[string][9]** the global coverage variable scope. (optional, default `this`)
+    -   `opts.coverageGlobalScopeFunc` **[boolean][10]** use an evaluated function to find coverageGlobalScope. (optional, default `true`)
     -   `opts.ignoreClassMethods` **[Array][11]** names of methods to ignore by default on classes. (optional, default `[]`)
     -   `opts.inputSourceMap` **[object][8]** the input source map, that maps the uninstrumented code back to the
         original code. (optional, default `undefined`)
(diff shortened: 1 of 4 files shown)


rhendric:
I have a very closely related problem that this PR doesn't quite solve, but could easily be adapted to solve: I want to deploy instrumented scripts in environments where [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) forbids `new Function('...')`. Would you be willing to make `coverageGlobalScope` replace the entire `(new Function('return this'))()` expression instead of just the `this`?

coreyfarrell:
@rhendric I cannot change the default but I can add another option.  The `new Function('...')` is used to escape the current context and make sure we run in the global context.  I'll be pushing a new commit that will allow `coverageGlobalScopeFunc: false` to override the default behavior,

PR #204, opened 2018-07-20
Support Typescipt or Flow static types

wzalazar:
This PR introduces:

Better compatibility when is instrumenting the code. Project with Typescript code has a lot of issues when flow can't instrument the code.

- Adds the option `staticType` with default value `flow`, it can be `flow` or `typescript`
- Unit test

### DIFF
--- a/packages/istanbul-lib-instrument/src/instrumenter.js
+++ b/packages/istanbul-lib-instrument/src/instrumenter.js
@@ -18,7 +18,8 @@ function defaultOpts() {
         produceSourceMap: false,
         ignoreClassMethods: [],
         sourceMapUrlCallback: null,
-        debug: false
+        debug: false,
+        staticType: 'flow'
     };
 }
 /**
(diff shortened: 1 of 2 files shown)


coreyfarrell:
I'm wondering if we could just use #205 to provide this?  I'm not too keen on the idea of having two options control plugins, it seems like it could become confusing.

wzalazar:
@coreyfarrell  

yeah, we can do it in the #205. regarding static types, it's important to know that plugin `flow` and `typescript` don't work together and we should somehow mention to the users. For this reason, I added this option separately

coreyfarrell:
Now that I've merged #205 I'm not sure what you want to do with this?  Maybe this could be solved by documenting the conflict?

wzalazar:
yes, mention it in the documentation is enough

JaKXz:
I think we should close this and have the doc PR separately. Thanks again for your contribution :)

PR #205, opened 2018-07-20
Add option plugins

wzalazar:
This PR introduces:

The ability to add more plugins for `@babel/parser`

- adds the option `plugins`
- unit test

### DIFF
--- a/packages/istanbul-lib-instrument/src/instrumenter.js
+++ b/packages/istanbul-lib-instrument/src/instrumenter.js
@@ -8,7 +8,7 @@ import traverse from '@babel/traverse';
 import generate from '@babel/generator';
 import programVisitor from './visitor';
 
-function defaultOpts() {
+export function defaultOpts() {
     return {
         coverageVariable: "__coverage__",
         preserveComments: false,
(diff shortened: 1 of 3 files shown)


packages/istanbul-lib-instrument/src/instrumenter.js
```diff
@@ -80,17 +82,18 @@ class Instrumenter {
         }
         filename = filename || String(new Date().getTime()) + '.js';
         const opts = this.opts;
+        const defaultPlugins = [
```
coreyfarrell: I'd prefer that `new Instrumenter({plugins: someList})` replace the default plugins.  The way I envision this:
1. Export `defaultOpts` (in src/index.js as well).
2. Move this defaultPlugins array to the `plugins` property inside `defaultOpts`.
3. If the caller wants to append default plugins they can import defaultOpts and use destructuring or Array.concat to setup opts.plugins.

I'm concerned that someone might need plugins in a specific order or may need to remove a plugin, the current method would not allow this.

wzalazar:
@coreyfarrell can you review the changes?

wzalazar:
@coreyfarrell awesome!!! thanks

coreyfarrell:
Thank you for your contribution and patience, I've merged this.

Issue #211, opened 2018-08-13
Mark HTML reports with eslint-disable


joshkel:
Istanbul's HTML coverage reports go in the `coverage` subdirectory of a project by default, which means they'll be picked up by a project's tooling (such as ESLint).

This can cause unwanted behavior, such as ESLint appearing to lock up from trying to process Istanbul's minified third-party JavaScript:

https://github.com/eslint/eslint/issues/10239#issuecomment-389288730

The correct fix is for projects that use Istanbul's HTML reports to correctly configure their tooling to exclude the `coverage` directory (for example, by adding `coverage/*` to `.eslintignore`).  However, it's easy for Istanbul to help accommodate tooling by adding appropriate comments to its coverage report files, telling tools to skip them.

PR #212, opened 2018-08-13
Mark HTML coverage report's files as eslint-ignore

joshkel:
Istanbul's HTML coverage reports go in the `coverage` subdirectory of a project by default, which means they'll be picked up by a project's tooling (such as ESLint).

This can cause unwanted behavior, such as ESLint appearing to lock up from trying to process Istanbul's minified third-party JavaScript:

https://github.com/eslint/eslint/issues/10239#issuecomment-389288730

The correct fix is for projects that use Istanbul's HTML reports to correctly configure their tooling to exclude the `coverage` directory (for example, by adding `coverage/*` to `.eslintignore`).  However, it's easy for Istanbul to help accommodate tooling by adding appropriate comments to its coverage report files, telling tools to skip them.

This PR adds support for optionally prepending headers when copying files and uses that to mark coverage JS files with `/* eslint-disable */`.

See #211.


coreyfarrell:
Merged, thank you for the contribution.

This actually explains why I've run into situations where eslint (or xo) would lock-up on me if the folder wasn't ignored.

PR #213, opened 2018-08-24
Support turning off node_modules default exclude via flag

hershmire:
This PR is the first step at resolving [NYC's issue #898](https://github.com/istanbuljs/nyc/issues/898) I raised. The idea is to be able to turn off the default exclusion of `**/node_modules/**` for users that need/want to support local `node_modules`. As mentioned in that issue, the `negateExclude` (i.e. `!**/node_modules/**`) doesn't work for cases when you still want your exclusion list to still exclude matches within your local `node_modules` directory. 

Related PRs for solving the above issue:
- https://github.com/istanbuljs/nyc/pull/912
- https://github.com/istanbuljs/babel-plugin-istanbul/pull/172

**NOTE:** Not sure if you also would want the `exportFunc.defaultExclude` list to take into account this option. For now, I'm not touching that so for this flag to work, at least one `exclude` must be present. Let me know if you would want this to also be included and I can make that edit. Thanks!

### DIFF
--- a/packages/test-exclude/index.js
+++ b/packages/test-exclude/index.js
@@ -14,7 +14,8 @@ class TestExclude {
                 relativePath: true,
                 configKey: null, // the key to load config from in package.json.
                 configPath: null, // optionally override requireMainFilename.
-                configFound: false
+                configFound: false,
+                excludeNodeModules: true
             },
             opts
         );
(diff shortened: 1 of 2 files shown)


JaKXz:
Sorry for the delayed reply @hershmire, would you mind rebasing this?

PR #225, opened 2018-10-08
Create LICENSE

warmans:
My company is strict about licencing (even for dev deps) and require an explicit licence file for vendored projects rather than simply naming the licence. I believe this is due to the possibility that the licence will change over time and without the file it's not clear what terms were agreed to. Anyway it would be much appreciated if you added the licence file.

Thanks.


coreyfarrell:
I'm not sure about this.  The git repository is not a package, the packages are in packages/*, and each one has it's own LICENSE file.  There is no 'npm install' command which will install this license file you've provided.  The LICENSE file you've provided doesn't appear to match that which is found in any of the packages.  I'll have to defer to @bcoe on this one.

warmans:
Ah OK I was not aware there were files in the directories. This PR is not needed in that case. Thank you for replying anyway.

PR #230, opened 2018-10-18
Fix for nested strategy when no files

chinchiheather:
This fixes an issue I was experiencing when using the nested summarizer and the coverage map had no files.

Previously it would throw an error
```
TypeError: other.contains is not a function
at Path.ancestorOf (/Users/heatherroberts/Desktop/Test Projects/jest/node_modules/istanbul-lib-report/lib/path.js:97:18)
```

### DIFF
--- a/packages/istanbul-lib-report/test/summarizer.test.js
+++ b/packages/istanbul-lib-report/test/summarizer.test.js
@@ -321,7 +321,15 @@ describe('summarizer', function() {
             assert.deepEqual(nodes, ['g:']);
         });
 
-        it('supports a list of files at top-level', function() {
+        it('handles getting root node name when empty coverage map', function() {
+          var map = coverage.createCoverageMap({}),
+              tree = fn(map),
+              root = tree.getRoot(),
+              rootNodeName = root.getRelativeName();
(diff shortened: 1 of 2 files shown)


bcoe:
@chinchiheather could I bother you to rebase this?

chinchiheather:
Hi @bcoe I have rebased

packages/istanbul-lib-report/test/summarizer.test.js
```diff
@@ -321,6 +321,14 @@ describe('summarizer', function() {
             assert.deepEqual(nodes, ['g:']);
         });
 
+        it('handles getting root node name when empty coverage map', function() {
```
JaKXz: my only minor nitpick here would be to describe the outcome. `handles` is good, but ideally this `does not crash` or something like that right?

chinchiheather:
I have made the change you requested @JaKXz

JaKXz:
Thank you @chinchiheather!

bcoe:
\o/ thank you @JaKXz @chinchiheather will work on cutting a release of istanbul soon.

Issue #224, opened 2018-09-28
Failed to write coverage reports


joe223:
```shell
Failed to write coverage reports:
ERROR: TypeError: Cannot read property 'decl' of undefined
STACK: TypeError: Cannot read property 'decl' of undefined
at /Users/joe/git/Oasis/node_modules/istanbul-reports/lib/lcovonly/index.js:32:38
at Array.forEach (<anonymous>)
at LcovOnlyReport.onDetail (/Users/joe/git/Oasis/node_modules/istanbul-reports/lib/lcovonly/index.js:30:28)
at LcovReport.(anonymous function) [as onDetail] (/Users/joe/git/Oasis/node_modules/istanbul-reports/lib/lcov/index.js:21:24)
at Visitor.(anonymous function) [as onDetail] (/Users/joe/git/Oasis/node_modules/istanbul-lib-report/lib/tree.js:34:30)
at ReportNode.Node.visit (/Users/joe/git/Oasis/node_modules/istanbul-lib-report/lib/tree.js:123:17)
at /Users/joe/git/Oasis/node_modules/istanbul-lib-report/lib/tree.js:116:23
at Array.forEach (<anonymous>)
at visitChildren (/Users/joe/git/Oasis/node_modules/istanbul-lib-report/lib/tree.js:115:32)
at ReportNode.Node.visit (/Users/joe/git/Oasis/node_modules/istanbul-lib-report/lib/tree.js:126:5)
```
> ./coverage/lcov.info
```
TN:
SF:unknown
```

> ./coverage/coverage-final.json
```json
		"s": {
			"0": 0,
			"1": 0,
			"2": 0,
			"3": 0,
			"4": 0,
			"5": 0,
			"6": 0,
			"7": 0,
			"8": 0,
			"9": 0,
			"10": 0,
			"11": 0,
			"12": 0
		},
		"f": {
			"0": 0
		},
		"b": {
			"0": [
				0,
				0
			]
		},
		"_coverageSchema": "332fd63041d2c1bcb487cc26dd0d5f7d97098a6c",
		"hash": "b2f30def596377432313c9168bee3ee9f2c11d04"
	}
}
```
What I can figure out is `var meta = functionMap[key]` will get `undefined` In this file https://github.com/istanbuljs/istanbuljs/blob/5405550c3868712b14fd8bfe0cbd6f2e7ac42279/packages/istanbul-reports/lib/lcovonly/index.js#L32

So，maybe we should add a conditional statement:
[...]

PR #253, opened 2018-12-11
bug fix for #224: failed to write coverage report (lcovonly)

Kosta-Github:
Replace `Object.keys(functions)` by `Object.keys(functionMap)` since inside of the loop `functionMap[key]` is used for lookup.

This is also the same behaviour as for the `cobertura` reporter here: https://github.com/istanbuljs/istanbuljs/blob/2857f990a85c9f276ec96f0e6969a33577be4ea7/packages/istanbul-reports/lib/cobertura/index.js#L98-L109
`Object.keys(fnMap)` is used for iteration and then `fnMap[k].decl.start.line` is used for accessing the data.

see: https://github.com/istanbuljs/istanbuljs/issues/224

### DIFF
--- a/packages/istanbul-reports/lib/lcovonly/index.js
+++ b/packages/istanbul-reports/lib/lcovonly/index.js
@@ -27,14 +27,14 @@ LcovOnlyReport.prototype.onDetail = function (node) {
     writer.println('TN:'); //no test name
     writer.println('SF:' + fc.path);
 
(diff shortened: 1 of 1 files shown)


bcoe:
@Kosta-Github this seems like a reasonable change, I wish I understood why there's a discrepancy between the two objects though -- since this only seems to happen periodically.

PR #255, opened 2018-12-12
 fix: pull in @loganfsmyth's SourceMap fixes

bcoe:
This pulls in @loganfsmyth's modifications to the remapping algorithm for SourceMaps.

I have a couple questions that I've left inline, please note that the tests that need to be modified reference the following:

```js
  // Original source
  // export class SimpleClass {
  //     hy() {
  //         console.log("Hy");
  //     }
  // }
  // Transpiled Source
  // "use strict";
  // var SimpleClass = (function () {
  //     function SimpleClass() {
  //     }
  //     SimpleClass.prototype.hy = function () {
  //         console.log("Hy");
  //     };
  //     return SimpleClass;
  // }());
  // exports.SimpleClass = SimpleClass;
```

CC: @loganfsmyth, @SimenB

### DIFF
--- a/packages/istanbul-lib-source-maps/lib/transformer.js
+++ b/packages/istanbul-lib-source-maps/lib/transformer.js
@@ -14,23 +14,87 @@ function isInvalidPosition (pos) {
 }
 
 /**
+ * AST ranges are inclusive for start positions and exclusive for end positions.
+ * Source maps are also logically ranges over text, though interacting with
+ * them is generally achieved by working with explicit positions.
+ *
+ * When finding the _end_ location of an AST item, the range behavior is
+ * important because what we're asking for is the _end_ of whatever range
+ * corresponds to the end location we seek.
+ *
+ * This boils down to the following steps, conceptually, though the source-map
+ * library doesn't expose primitives to do this nicely:
+ *
+ * 1. Find the range on the generated file that ends at, or exclusively
+ *    contains the end position of the AST node.
+ * 2. Find the range on the original file that corresponds to
+ *    that generated range.
+ * 3. Find the _end_ location of that original range.
+ */
+function originalEndPositionFor(sourceMap, generatedEnd) {
+    // Given the generated location, find the original location of the mapping
+    // that corresponds to a range on the generated file that overlaps the
+    // generated file end location. Note however that this position on its
+    // own is not useful because it is the position of the _start_ of the range
+    // on the original file, and we want the _end_ of the range.
+    var beforeEndMapping = sourceMap.originalPositionFor({
+        line: generatedEnd.line,
+        column: generatedEnd.column - 1,
+    });
+    if (beforeEndMapping.source === null) {
+        return null;
+    }
+
+    // Convert that original position back to a generated one, with a bump
+    // to the right, and a rightward bias. Since 'generatedPositionFor' searches
+    // for mappings in the original-order sorted list, this will find the
+    // mapping that corresponds to the one immediately after the
+    // beforeEndMapping mapping.
+    var afterEndMapping = sourceMap.generatedPositionFor({
+        source: beforeEndMapping.source,
+        line: beforeEndMapping.line,
+        column: beforeEndMapping.column + 1,
+        bias: 2
+    });
+    if (
+        // If this is null, it means that we've hit the end of the file,
+        // so we can use Infinity as the end column.
+        afterEndMapping.line === null ||
+
+        // If these don't match, it means that the call to
+        // 'generatedPositionFor' didn't find any other original mappings on
+        // the line we gave, so consider the binding to extend to infinity.
+        sourceMap.originalPositionFor(afterEndMapping).line !== beforeEndMapping.line
+    ) {
+        return {
+            source: beforeEndMapping.source,
+            line: beforeEndMapping.line,
+            column: Infinity,
+        };
+    }
+
+    // Convert the end mapping into the real original position.
+    return sourceMap.originalPositionFor(afterEndMapping);
+}
+
+/**
  * determines the original position for a given location
  * @param  {SourceMapConsumer} sourceMap the source map
- * @param  {Object} location the original location Object
+ * @param  {Object} generatedLocation the original location Object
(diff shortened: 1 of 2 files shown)


packages/istanbul-lib-source-maps/test/map-store.test.js
```diff
@@ -36,7 +36,7 @@ describe('map store', function () {
         },
         "end": {
           "line": 4,
-          "column": 5
+          "column": Infinity
```
bcoe: @loganfsmyth is it weird that line 4 (in a 5 line file) extends to the column position `Infinity`? Your comment indicates that this edge-case should only occur when we've hit the end of the file.

packages/istanbul-lib-source-maps/test/map-store.test.js
```diff
@@ -79,8 +79,8 @@ describe('map store', function () {
             "column": 0
           },
           "end": {
-            "line": null,
-            "column": -1
+            "line": 1,
+            "column": 13
```
bcoe: It seemed a bit weird that the `decl` mapping for `(anonymous_1)` suddenly started mapping to something; what is `(anonymous_1)`? I think this might be referring to `export class` as if it represents a function?

packages/istanbul-lib-source-maps/test/map-store.test.js
```diff
@@ -102,8 +102,8 @@ describe('map store', function () {
             "column": 4
           },
           "end": {
-            "line": null,
-            "column": -1
+            "line": 2,
```
bcoe: similarly it seemed weird that `(anonymous_3)` had a different `decl` which includes the entire function body, as `decl` is supposed to only include the class declaration I believe, so: `function foo ()`.

This could be the fault of Istanbul though, and how code is transpiled?

```js
  // var SimpleClass = (function () {
  //     function SimpleClass() {
  //     }
  //     SimpleClass.prototype.hy = function () {
  //         console.log("Hy");
  //     };
  //     return SimpleClass;
  // }());
```

bcoe:
@coreyfarrell would love your eye balls on this.

packages/istanbul-lib-source-maps/lib/transformer.js
```diff
@@ -13,24 +13,88 @@ function isInvalidPosition (pos) {
...
+    // Convert that original position back to a generated one, with a bump
+    // to the right, and a rightward bias. Since 'generatedPositionFor' searches
+    // for mappings in the original-order sorted list, this will find the
+    // mapping that corresponds to the one immediately after the
+    // beforeEndMapping mapping.
+    var afterEndMapping = sourceMap.generatedPositionFor({
+        source: beforeEndMapping.source,
+        line: beforeEndMapping.line,
+        column: beforeEndMapping.column + 1,
+        bias: 2
```
loganfsmyth: I had this because the existing code does it already, but FYI the bias values are also exposed as constants:
```
bias: SourceMapConsumer.LEAST_UPPER_BOUND
```
if you want.

packages/istanbul-lib-source-maps/test/map-store.test.js
loganfsmyth: The difficulty is that sourcemap mappings don't actually contain the end position data, so all you know is that it extends to the end of a given line, with no sense of what column that might be. You'd have to fetch the source content and split it by line to know the actual column offset.

I agree though that it may well require some downstream validation. I guess you could always just do `Number.MAX_SAFE_INTEGER` if you wanted to avoid dangers with possible `Infinity` values or something, but hard to say if that's actually any less likely to have issues.

packages/istanbul-lib-source-maps/test/map-store.test.js
loganfsmyth: I'm having a hard time figuring out how this data blob works, so not sure I can add anything.

packages/istanbul-lib-source-maps/test/map-store.test.js
coreyfarrell: I haven't done a full review yet (I have other commitments today), but I just wanted to comment that I did check `Infinity` against the second argument to `String.slice`, `String.substr` and `String.substring`.  All worked as expected.

I'm could be convinced if someone could show a way that `Infinity` could break something but I'd prefer we did not use `Number.MAX_SAFE_INTEGER`.  `Infinity` is clearly not a specific number where `Number.MAX_SAFE_INTEGER` is.

bcoe: the fact that `Infinity` behaves nicely with the String helpers is good to know. We're talking about making this a breaking change any ways, and I'm feeling more and more confident about swapping out what were already undefined ranges with `Infinity1.

packages/istanbul-lib-source-maps/test/map-store.test.js
bcoe: @loganfsmyth is it possible we're treating the `export` or `class` snippet as a `function` token type in the AST, and this is what `(anonymous_1)` refers to? The new offsets seem better than `null`, and `-1` ... which seems like a bug with a test around it, no?

@coreyfarrell does this set off any alarm bells for you?

packages/istanbul-lib-source-maps/test/map-store.test.js
coreyfarrell: Granted `undefined` also works with String helpers but does not work with comparisons.
```js
assert(5 < Infinity); // pass
assert(5 < undefined); // fail
```

So I think this is 1) an improvement and 2) appropriate to flag as a breaking change just in case.

packages/istanbul-lib-source-maps/test/map-store.test.js
coreyfarrell: The `(anonymous_1)` is referring to the IIFE.

So we've instrumented the transpiled source rather than instrumenting the original source.  I'm not fully sure how this is supposed to work, on my own projects I put instrumentation before other babel transformations.  I realize that could probably be a problem with TypeScript (the code added by instrumentation may not pass TypeScript checks).

PR #257, opened 2018-12-19
fix: correct variable name in source-map transform

SimenB:
A linter should probably have caught this


bcoe:
@loganfsmyth any thoughts on a way to exercise this line? @SimenB if you saw fit to plop in a linter (which I think we should have setup on the other packages perhaps?) it would be amazing.

SimenB:
Can this be merged and released in the meantime, the current prerelease is pretty broken due to this 🙂 

I can send a PR for some linting, sure

PR #258, opened 2018-12-20
add eslint with prettier

SimenB:
As requested in #257.

I went with eslint recommended rules, and prettier. I have left the violations that was not autofixed (so all changed code is autofixed).

image

You can see the bug from #257 in the middle 🙂 

Could consider e.g. `prefer-arrow` and `template-string` rules, which are both autofixable, but left that out.

I also deleted some old jshint stuff.


bcoe:
closing in favor of https://github.com/istanbuljs/istanbuljs/pull/261

PR #268, opened 2019-01-10
chore: Update dependencies.

coreyfarrell:
Update all dependencies except source-map which requires node.js 8.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,23 +5,23 @@
 	"requires": true,
 	"dependencies": {
 		"@babel/code-frame": {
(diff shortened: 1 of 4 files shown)


JaKXz:
@coreyfarrell I'll let you merge when you are ready! thanks for doing these

coreyfarrell:
I'm going to try again upgrading lerna.  It's possible I just didn't properly regenerate package-lock.json the first time.

coreyfarrell:
OK apparently lerna@3 breaks package-lock.json for node.js 8.  Will have to deal with this later.

PR #272, opened 2019-01-14
build: update travis config to include windows

JaKXz:
and latest LTS node versions


package.json
```diff
@@ -29,14 +29,14 @@
...
-    "coveralls": "^3.0.2",
     "chai": "^4.1.2",
     "clone": "^2.1.1",
+    "coveralls": "^3.0.2",
     "cross-env": "^5.2.0",
     "debug": "^4.1.1",
     "documentation": "^7.1.0",
     "eslint": "^5.10.0",
-    "eslint-config-prettier": "^3.3.0",
+    "eslint-config-prettier": "^3.4.0",
```
coreyfarrell: The default for git on Windows is to convert line endings to `\r\n` in working copies, so I think `prettier/prettier` is throwing errors because of this.

JaKXz: You're right... that's why we had these lines in our appveyor.yml:

```yml
init:
  - git config --global core.autocrlf false
```

coreyfarrell: Is it possible to tell prettier to allow CRLF?  I'd prefer that so we verify the best compatibility (since core.autocrlf=false is not default).

PR #274, opened 2019-01-15
chore: prettierignore files modified on release

JaKXz:
Something we forgot in #261 was to make sure our automated tools (semantic-version etc) don't get messed with by prettier on release


JaKXz:
We should - Webstorm told me that all the `package*.json` files in sub directories would be ignored. 

And I only saw CHANGELOG.md files in the sub directories. But I'll make them consistent.

PR #275, opened 2019-01-16
Add support to exclude files on writing reports

henriquemotaesteves:
This pull request is an attempt to close the issue https://github.com/istanbuljs/nyc/issues/956.

### DIFF
--- a/packages/istanbul-api/lib/reporter.js
+++ b/packages/istanbul-api/lib/reporter.js
@@ -6,7 +6,8 @@ var path = require('path'),
     configuration = require('./config'),
     inputError = require('./input-error'),
     libReport = require('istanbul-lib-report'),
-    libReports = require('istanbul-reports');
+    libReports = require('istanbul-reports'),
+    minimatch = require('minimatch');
 
 function Reporter(cfg, opts) {
     opts = opts || {};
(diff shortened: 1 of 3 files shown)


packages/istanbul-api/test/reporter.test.js
```diff
@@ -0,0 +1,106 @@
...
+                    coverage.createFileCoverage('/x/b/c.js')
+                );
+                coverageMap.addFileCoverage(
+                    coverage.createFileCoverage('/d/e/f.js')
+                );
+                coverageMap.addFileCoverage(
+                    coverage.createFileCoverage('/g/x/i.js')
+                );
+
+                it('does not exclude files from reports', function(cb) {
```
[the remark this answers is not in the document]
henriquemotaesteves: If I understood correctly, the requested change was to rename `cb` to `done`. Done.

JaKXz:
@coreyfarrell I think this makes sense as a non-breaking feature, but I'll leave that call to you :)

packages/istanbul-api/lib/reporter.js
```diff
@@ -6,7 +6,8 @@ var path = require('path'),
     configuration = require('./config'),
     inputError = require('./input-error'),
     libReport = require('istanbul-lib-report'),
-    libReports = require('istanbul-reports');
+    libReports = require('istanbul-reports'),
+    minimatch = require('minimatch');
```
coreyfarrell: `minimatch` needs to be added to packages/istanbul-api/package.json as a `dependency`.

henriquemotaesteves: Done!

packages/istanbul-api/test/reporter.test.js
```diff
@@ -0,0 +1,106 @@
+/* globals context, describe, it */
+
+var assert = require('chai').assert,
+    configuration = require('../lib/config'),
+    coverage = require('../../istanbul-lib-coverage'),
```
coreyfarrell: Sorry I didn't catch this previously, could you change this to simply `coverage = require('istanbul-lib-coverage')`?  This way the module is imported from `node_modules` instead of directly from the relative directory.

henriquemotaesteves: Done!

PR #282, opened 2019-01-28
feat: add nyc-config-typescript

JaKXz:
Kind of a carbon copy of the typescript tutorial and `nyc-config-babel` to try to help with the typescript issues we're seeing...


packages/nyc-config-typescript/index.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extension": [
+    ".ts",
+    ".tsx"
+  ],
+  "exclude": [
```
coreyfarrell: `exclude` does not get merged with the default config (unfortunately).  So I think you should add the default excludes from test-exclude:
```js
    'coverage/**',
    'packages/*/test/**',
    'test/**',
    'test{,-*}.js',
    '**/*{.,-}test.js',
    '**/__tests__/**',
    '**/node_modules/**',
    '**/babel.config.js'
```

packages/nyc-config-typescript/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@istanbuljs/nyc-config-typescript",
+  "version": "1.0.0",
```
coreyfarrell: Do you want to start immediately at 1.0.0?  Many changes to index.json would be considered semver-major, maybe start with 0.1.0 until we get some feedback?  That way semver-major changes can go into 0.2.0, 0.3.0 then 1.0.0 can be stable/longer lived?

packages/nyc-config-typescript/README.md
```diff
@@ -0,0 +1,55 @@
+# nyc-config-typescript
+
+Handy default configuration for instrumenting your babel-backed
+project with test coverage using [nyc](https://github.com/istanbuljs/nyc).
+
+First install the dependencies:
+
+`npm i source-map-support ts-node @istanbuljs/nyc-config-typescript --save-dev`
```
coreyfarrell: Should this include `nyc`?

JaKXz:
Thanks @coreyfarrell I think I addressed all your comments. Let me know what you think.

packages/nyc-config-typescript/index.json
```diff
@@ -0,0 +1,17 @@
+{
+  "cache": false,
```
coreyfarrell: I'm don't know why you would want to disable nyc cache when using ts-node but I also don't know enough to object to it.

JaKXz: I've seen some issues pop up as a result of `cache: true`, so, just for the tutorial/starting package it's probably better. It can of course be overridden in `.nycrc` etc

PR #284, opened 2019-01-28
build: fix travis config

JaKXz:
Windows tests can come later

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: node_js
+os:
+  - osx
+  - linux
 node_js:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,15 +1,12 @@
-sudo: false
 language: node_js
 node_js:
```
JaKXz: the only question is do we want:

```yml
os:
  - linux
  - osx
```

or would that be overkill?

coreyfarrell: I'd say lets add osx.

PR #287, opened 2019-02-04
Map unique files once, regardless of path separator

erikbarke:
The paths "/path/to/file.js" and "\path\to\file.js" points to the same file on Windows, this PR makes `istanbuljs` treat them as one file by using `_path_to_file.js` as a key internally when transforming coverage maps.

Long story short, this makes `istanbuljs` behave well without reporting duplicates, regardless of input.

### DIFF
--- a/packages/istanbul-lib-source-maps/lib/transformer.js
+++ b/packages/istanbul-lib-source-maps/lib/transformer.js
@@ -234,12 +234,17 @@ SourceMapTransformer.prototype.processFile = function(
 SourceMapTransformer.prototype.transform = function(coverageMap) {
     var that = this,
         finder = this.finder,
-        output = {},
+        uniqueFiles = {},
+        key,
         getMappedCoverage = function(file) {
-            if (!output[file]) {
-                output[file] = new MappedCoverage(file);
+            key = getUniqueKey(file);
+            if (!uniqueFiles[key]) {
+                uniqueFiles[key] = {
+                    file: file,
+                    mappedCoverage: new MappedCoverage(file)
+                };
             }
-            return output[file];
+            return uniqueFiles[key].mappedCoverage;
         };
 
     coverageMap.files().forEach(function(file) {
@@ -248,7 +253,10 @@ SourceMapTransformer.prototype.transform = function(coverageMap) {
             changed;
 
         if (!sourceMap) {
-            output[file] = fc;
+            uniqueFiles[getUniqueKey(file)] = {
+                file: file,
+                mappedCoverage: fc
+            };
             return;
         }
 
@@ -257,8 +265,22 @@ SourceMapTransformer.prototype.transform = function(coverageMap) {
             debug('File [' + file + '] ignored, nothing could be mapped');
         }
     });
-    return libCoverage.createCoverageMap(output);
+    return libCoverage.createCoverageMap(getOutput(uniqueFiles));
 };
+
+function getUniqueKey(path) {
+    return path.replace(/[\\/]/g, '_');
+}
+
+function getOutput(cache) {
+    var output = {},
+        item;
+    Object.keys(cache).forEach(function(key) {
+        item = cache[key];
+        output[item.file] = item.mappedCoverage;
+    });
+    return output;
+}
 
 module.exports = {
     create: function(finder, opts) {
--- a/packages/istanbul-lib-source-maps/lib/transformer.js
+++ b/packages/istanbul-lib-source-maps/lib/transformer.js
@@ -273,13 +273,12 @@ function getUniqueKey(path) {
 }
 
 function getOutput(cache) {
-    var output = {},
-        item;
-    Object.keys(cache).forEach(function(key) {
-        item = cache[key];
-        output[item.file] = item.mappedCoverage;
-    });
-    return output;
+    return Object.keys(cache).reduce((output, key) => {
+        const item = cache[key];
+        return Object.assign(output, {
+            [item.file]: item.mappedCoverage
+        });
+    }, {});
 }
 
 module.exports = {
--- a/packages/istanbul-lib-source-maps/test/transformer.test.js
+++ b/packages/istanbul-lib-source-maps/test/transformer.test.js
@@ -3,17 +3,8 @@ var assert = require('chai').assert,
     isWindows = require('is-windows'),
     createMap = require('istanbul-lib-coverage').createCoverageMap,
     SMC = require('source-map').SourceMapConsumer,
-    createTransformer = require('../lib/transformer').create;
-
-function createData() {
-    var sourceMap = {
-        version: 3,
-        sources: ['file.js'],
-        mappings: ';AAAa,mBAAW,GAAG,MAAM,CAAC;AACrB,kBAAU,GAAG,yBAAyB,CAAC'
-    };
-
-    var coverageData = {
-        path: '/path/to/file.js',
+    createTransformer = require('../lib/transformer').create,
+    coverageData = {
         statementMap: {
             '0': {
                 start: { line: 2, column: 0 },
@@ -35,38 +26,24 @@ function createData() {
         b: {}
     };
 
+function createData() {
+    var data = Object.assign({}, coverageData);
+    data.path = '/path/to/file.js';
     return {
-        sourceMap: sourceMap,
-        coverageData: coverageData
+        sourceMap: {
+            version: 3,
+            sources: ['file.js'],
+            mappings: ';AAAa,mBAAW,GAAG,MAAM,CAAC;AACrB,kBAAU,GAAG,yBAAyB,CAAC'
+        },
+        coverageData: data
     };
 }
 
 function createDataBackslash() {
-    var coverageData = {
-        path: '\\path\\to\\file.js',
-        statementMap: {
-            '0': {
-                start: { line: 2, column: 0 },
-                end: { line: 2, column: 29 }
-            },
-            '1': {
-                start: { line: 3, column: 0 },
-                end: { line: 3, column: 47 }
-            }
-        },
-        fnMap: {},
-        branchMap: {},
-        s: {
-            '0': 0,
-            '1': 0,
-            '2': 0
-        },
-        f: {},
-        b: {}
-    };
-
+    var data = Object.assign({}, coverageData);
+    data.path = '\\path\\to\\file.js';
     return {
-        coverageData: coverageData
+        coverageData: data
     };
 }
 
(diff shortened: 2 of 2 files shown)


erikbarke:
I agree, having 15-20 lines of (much repeated) setup in each test with one or two asserts isn't exactly optimal...

Without having delved too deep into the code, I have a gut feeling the setup could be trimmed down by mocking the stuff from the `istanbul-lib-coverage` package, because then the tests wouldn't need all the 'real' data added with `coverageMap.addFileCoverage` etc.

From the looks of it what should be asserted in the test is, for example, has `coverageMap.files()` been called? Has `coverageMap.fileCoverageFor` been called n times? Has `libCoverage.createCoverageMap` been called with '{foo: 'bar'}' and so on instead of relying on the actual code in the `istanbul-lib-coverage` package.

Just my two cents ☺️

packages/istanbul-lib-source-maps/lib/transformer.js
```diff
@@ -257,9 +265,23 @@ SourceMapTransformer.prototype.transform = function(coverageMap) {
...
 };
 
+function getUniqueKey(path) {
+    return path.replace(/[\\/]/g, '_');
+}
+
+function getOutput(cache) {
+    var output = {},
+        item;
+    Object.keys(cache).forEach(function(key) {
```
JaKXz: nit: this could be `Object.keys(cache).reduce(...)` and I think it would read a lot nicer.

packages/istanbul-lib-source-maps/test/transformer.test.js
```diff
@@ -53,6 +41,35 @@ function createData() {
     };
 }
 
+function createDataBackslash() {
+    var coverageData = {
+        path: '\\path\\to\\file.js',
```
JaKXz: For now we could use `Object.assign` here to save some of the repetition. As far as I can tell this is the only line that changes b/w these two fns so we could have a common `const` that is shared in a helper file or just at the top of this one.

packages/istanbul-lib-source-maps/test/transformer.test.js
```diff
@@ -80,4 +97,29 @@ describe('transformer', function() {
             }
         });
     });
+
+    it('maps each file only once, /path/to/file.js and \\path\\to\\file.js are the same file', function() {
+        if (isWindows()) {
+            return this.skip();
```
JaKXz: I'm just asking because I don't know: should this be skipped on windows? Don't windows paths have `\\`s in them?

packages/istanbul-lib-source-maps/lib/transformer.js
```diff
@@ -248,7 +253,10 @@ SourceMapTransformer.prototype.transform = function(coverageMap) {
             changed;
 
         if (!sourceMap) {
-            output[file] = fc;
+            uniqueFiles[getUniqueKey(file)] = {
```
JaKXz: My memory is a bit rusty here, what is the difference between this and `getMappedCoverage`? It looks like there's duplication that can be cleaned up but I could be wrong...

cc @coreyfarrell

packages/istanbul-lib-source-maps/lib/transformer.js
erikbarke: I'm not sure I'm following here, the code loops over the `cache` keys to populate the `output` object with data from each cached item, it's not accumulating a single sum?

packages/istanbul-lib-source-maps/test/transformer.test.js
erikbarke: A classic case of cargo cult programming, I don't know if it's needed or not,  I included it because it's also used in the 'maps statement locations' test 😉The test data includes windows style data, it doesn't rely on executing on windows, I'll try running the test without it!

packages/istanbul-lib-source-maps/lib/transformer.js
JaKXz: It could look like this:

```js
function getOutput(cache) {
  return Object.keys(cache).reduce((output, key) => {
    const item = cache[key];
    return Object.assign(output, {
      [item.file]: item.mappedCoverage
    });
  }, {});
}
```

If you think about it, it is accumulating (or transforming, or "reducing") the values of the cache object to another object. So really everything can be a reduce 😉 but like I said, this is just a stylistic nit.

I do want the `var output, item` comma declaration changed, I'm surprised prettier didn't catch that!

packages/istanbul-lib-source-maps/lib/transformer.js
erikbarke: The `getMappedCoverage` function is used if you have source maps and want to get coverage for the original code. If there are no source maps, just return the file coverage as is, so both are needed I guess.

packages/istanbul-lib-source-maps/test/transformer.test.js
erikbarke: Yeah, I'll try to reduce the amount of duplicated test data!

packages/istanbul-lib-source-maps/lib/transformer.js
erikbarke: Done!

packages/istanbul-lib-source-maps/test/transformer.test.js
erikbarke: Yep, I tried, this check is necessary!

packages/istanbul-lib-source-maps/test/transformer.test.js
erikbarke: Done!

erikbarke: Btw, I used a `var` since everything else is declared in a single `var` at the top, maybe the whole file should be updated to use `const`, `let` etc?

coreyfarrell: I haven't had a chance to look at your updated PR yet but I'd say leave it on `var` for this PR.  Switching to `const` / `let` would be a nice follow-up.  It would be a 'mass change' that could be easily reviewed on it's own but would bulk this PR.

JaKXz: @coreyfarrell I think it would be worth making the `const` `let` change as we go, instead of leaving it as tech debt to be cleaned up later. Gradual improvements as PRs are made make the most sense IMHO

coreyfarrell: Ah yes using `const` / `let` for new or modified code does make sense.  And on second look this file is not so large so I guess updating this whole file wouldn't be a big deal.

packages/istanbul-lib-source-maps/test/transformer.test.js
```diff
@@ -47,9 +26,24 @@ function createData() {
         b: {}
     };
 
+function createData() {
```
JaKXz: My intention with my comment was to move away from these one-shot factory functions in favour of having one base object that's extended.

packages/istanbul-lib-source-maps/test/transformer.test.js
```diff
@@ -47,9 +26,24 @@ function createData() {
...
+    var data = Object.assign({}, coverageData);
+    data.path = '/path/to/file.js';
```
JaKXz: This should be part of the `Object.assign`

packages/istanbul-lib-source-maps/test/transformer.test.js
```diff
@@ -80,4 +74,29 @@ describe('transformer', function() {
...
         });
     });
+
+    it('maps each file only once, /path/to/file.js and \\path\\to\\file.js are the same file', function() {
+        if (isWindows()) {
+            return this.skip();
+        }
+
+        var coverageMap = createMap({}),
+            testDataSlash = createData(),
```
JaKXz: Ideally we could just have `const testDataSlash = Object.assign({}, coverageData, { stuff: 'we need for this test' })` and so on.

packages/istanbul-lib-source-maps/test/transformer.test.js
JaKXz: @coreyfarrell @bcoe thoughts?

erikbarke: These lines seem to the relevant ones:

```
  transformer
    1) maps statement locations
    - maps each file only once, /path/to/file.js and \path\to\file.js are the same file
  4 passing (15ms)
  1 pending
  1 failing
  1) transformer
       maps statement locations:
     TypeError: Cannot read property 'statementMap' of undefined
      at Context.<anonymous> (test\transformer.test.js:85:57)
npm ERR! Test failed.  See above for more details.
lerna ERR! npm test exited ENOENT in 'istanbul-lib-source-maps'
```

erikbarke: The build is for Node.js v8.15.0 and it also complains about "ES6 feature [undefined] is not available in this environment" right after listing coverage stats.

packages/istanbul-lib-source-maps/test/transformer.test.js
erikbarke: Ok, the test file is converted to `const` 👍

packages/istanbul-lib-source-maps/test/transformer.test.js
erikbarke: Updated now, the tests are a lot easier to read now imho ☺️

JaKXz: 🙌

packages/istanbul-lib-source-maps/test/transformer.test.js
erikbarke: I can reproduce this error in Parallels on my mac, I'll try to figure it out.

erikbarke: It's a windows path issue, the input path '/path/to/file.js' comes back as 'c:\path\to\files.js' on windows which makes `mapped.data[testDataSlash.coverageData.path]` undefined...

I've tried resolving this by creating test data paths manually (using `__dirname`, `path.join` and a regex replace), which makes the paths (and slashes) look and behave consistently in the tests on all platforms ☺️

Using `path.resolve` and `path.normalize` didn't work because the paths get converted to what the current os expects, not what the tests expect.

coreyfarrell: Have you tried `path.posix.join` and such?

https://nodejs.org/dist/latest-v6.x/docs/api/path.html#path_path_posix

erikbarke: @coreyfarrell, it looks much better now, short and elegant!

JaKXz:
This is 🔥 @erikbarke! Great work.

packages/istanbul-lib-source-maps/lib/transformer.js
JaKXz: @coreyfarrell / @bcoe this is my only question left for this PR so if you feel this is fine then feel free to :shipit:

coreyfarrell: This looks good to me.

coreyfarrell:
Thanks for this fix @erikbarke!

PR #289, opened 2019-02-07
Update sorter.js

Michael03:
Removed unused  cols variable as loadData has no parameters.


coreyfarrell:
It looks like the `loadColumns()` function has side-effects, so even if we don't use the return value we probably still have to call it.

Michael03:
Updated to add loadColumns back in. 👍

coreyfarrell:
Thanks for the patch, merged.

PR #298, opened 2019-02-18
fix: Honor istanbul ignore next hints placed before export statement.

### LINKED ISSUE #297: Coverage ignore hints not working for export
After upgrading to Babel 7, our ignore hints are no longer working like they used to. 
In my case a lot of it (the hints not working) seems to be related to ignoring the export directly.

This does not get ignored:
```
/* istanbul ignore next */
export default function MyFunction () {
}
```

This gets ignored fine:
```
/* istanbul ignore next */
function MyFunction () {
}
export default MyFunction;
```

Possible related issues:
https://github.com/istanbuljs/istanbuljs/issues/157
https://github.com/istanbuljs/istanbuljs/issues/162
https://github.com/istanbuljs/babel-plugin-istanbul/issues/150

coreyfarrell:
Fixes #297

### DIFF
--- a/packages/istanbul-lib-instrument/src/visitor.js
+++ b/packages/istanbul-lib-instrument/src/visitor.js
@@ -484,6 +484,8 @@ const codeVisitor = {
     ArrowFunctionExpression: entries(convertArrowExpression, coverFunction),
     AssignmentPattern: entries(coverAssignmentPattern),
     BlockStatement: entries(), // ignore processing only
+    ExportDefaultDeclaration: entries(), // ignore processing only
+    ExportNamedDeclaration: entries(), // ignore processing only
     ClassMethod: entries(coverFunction),
     ClassDeclaration: entries(parenthesizedExpressionProp('superClass')),
     ObjectMethod: entries(coverFunction),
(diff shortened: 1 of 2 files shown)


coreyfarrell:
prettier :angry:

JaKXz:
@coreyfarrell protip: I use `npx pretty-quick --staged` when committing.

coreyfarrell:
@JaKXz normally I wouldn't do a change without running `npm test` which would run eslint.  In this case it took me a few tries to get the test working so I ran tests against istanbul-lib-instrument only - which bypassed the `posttest` in the root package.json.

Really my frustration with prettier is mostly just personal preference, complying with prettier style is difficult for me.

JaKXz:
I've had this discussion _at [way too much] length_ with colleagues, so I mean this with the best possible intent: I try not to worry too much about "complying" with the style. I just make my changes, stage them, and let `pretty-quick` take care of compliance. 

FWIW it was extremely liberating once I thought about it this way, I too had similar apprehensions at first.

coreyfarrell:
As long as I remember to run the full `npm test` instead of short-cutting it eslint tells me what I did wrong then I just run `npx eslint . --fix` (the test failure tells me to do this so this is what I remember).

I don't really have apprehensions about prettier, it's just annoying to me.  Personally I don't find that prettier produces more readable code, I especially dislike the rules about line length.

PR #303, opened 2019-02-19
BREAKING: Update to source-map@0.7.x, drop node.js 6.

coreyfarrell:
This is not yet working but I wanted to post it anyways.  This updates source-map, but it is an API breaking change to istanbul-lib-source-map.  The constructor in `source-map@0.6.x` returned a SourceMapConsumer object, the same for `source-map@0.7.x` returns a Promise which resolves to the SourceMapConsumer object.  I've updated the tests for istanbul-lib-source-map, including a couple places where the new source-map module produces slightly different results (not sure if it's right or wrong).

This change has a ripple effect of converting functions to use async/await.  `istanbul-api` needs to be updated to support these new API's, this will likely require converting parts of that module to async functions.

One question I have is how far we should go with this.  Should we use async only where necessary for the new source-map module, or should we use async/await across the board.  For example we currently use sync versions of some `fs` functions - do we change this?

Marking this PR as a draft since it should not be merged (we're not dropping node.js 6 until April).  It'll take some time to work out all the issues for this update so I just want to share code early.


JaKXz:
It would be cool if we could leverage (or even write) codemods that did some of this work. e.g. changing all `Object.assign`s to the spread operator and so on

coreyfarrell:
@JaKXz I'm not aware of anything that'll do that automatically but I plan to push #304 ahead of this (I'll rebase this once that is merged).  Will probably do a similar ES2015 patch for `istanbul-api` before I touch it here.

coreyfarrell:
Upgrading source-map is simply not possible, we need to use it in exit handlers which cannot be async. :-1:

PR #304, opened 2019-02-20
chore: ES2015ify istanbul-lib-source-maps.

coreyfarrell:
Use node.js 6 features.  This also makes a few changes requested by `xo` rules where they don't conflict with `prettier`.  If we stick with prettier I would like to start adding some of the `xo` rules to our `.eslintrc`.  For example [unicorn/no-new-buffer](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/no-new-buffer.md) raised an error about `new Buffer`.


packages/istanbul-lib-source-maps/lib/map-store.js
```diff
@@ -4,169 +4,186 @@
...
-                type: 'encoded',
-                data: sourceMapUrl.substring(pos + b64.length)
+                type: 'object',
+                data: sourceMap
             };
         } else {
-            debug('Unable to interpret source map URL: ' + sourceMapUrl);
+            debug(
+                'Invalid source map object:' +
+                    JSON.stringify(sourceMap, null, 2)
```
JaKXz: We could use string interpolation here

packages/istanbul-lib-source-maps/lib/pathutils.js
```diff
@@ -1,19 +1,23 @@
...
-        if (path.isAbsolute) {
-            return path.isAbsolute(p);
-        }
-        return path.resolve(p) === path.normalize(p);
-    };
+const path = require('path');
 
-exports.isAbsolute = isAbsolute;
+function isAbsolute(p) {
+    if (path.isAbsolute) {
```
JaKXz: This seems like check for old versions of node, do we still need it?

packages/istanbul-lib-source-maps/lib/source-store.js
```diff
@@ -1,78 +1,80 @@
...
-function SourceStore(/*opts*/) {}
-
-SourceStore.prototype.registerSource = function(/* filePath, sourceText */) {
-    throw new Error('registerSource must be overridden');
-};
-
-SourceStore.prototype.getSource = function(/* filePath */) {
-    throw new Error('getSource must be overridden');
-};
+class SourceStore {
```
JaKXz: If this is meant to be an abstract class, we should have some sort of enforcement:

```js
class SourceStore {
  constructor() {
    if (this.constructor === SourceStore) throw new Error('Abstract class SourceStore must be extended');
  }
}
```

packages/istanbul-lib-source-maps/lib/transformer.js
```diff
@@ -148,141 +151,138 @@ function getMapping(sourceMap, generatedLocation, origFile) {
     };
 }
 
-function SourceMapTransformer(finder, opts) {
-    opts = opts || {};
-    this.finder = finder;
-    this.baseDir = opts.baseDir || process.cwd();
+function getUniqueKey(path) {
```
JaKXz: nitpick: this is sort of a big file, is there any separation we can do? e.g. moving out utility methods or at least moving them to the bottom? My personal preference is having the thing that the file is named after exported at the very top (with the `export default` keywords) but that may be N/A here.

JaKXz:
@coreyfarrell I think adding the unicorn plugin to the `.eslintrc` would be a good way to go, but, IME some of those rules are extremely strict or for a very specific use case... https://github.com/mysticatea/eslint-plugin-node#readme is a better fit IMHO.

packages/istanbul-lib-source-maps/lib/transformer.js
coreyfarrell: I've moved to a couple separate files.  Personally my preference is against hoisting, so declaring a function before it is used.  Pulling from a separate file is fine as well.

packages/istanbul-lib-source-maps/lib/source-store.js
coreyfarrell: Not sure what happened to my comment about this.  I've updated this class to be completely empty, created a noop `MemoryStore.dispose()`.  It's not exported and is only used as a shared base class.  I'm not sure it needs to exist but for now I figure keep the prototype chain the same.

packages/istanbul-lib-source-maps/lib/transform-utils.js
```diff
@@ -0,0 +1,20 @@
+/*
+ Copyright 2015, Yahoo Inc.
+ Copyrights licensed under the New BSD License. See the accompanying LICENSE file for terms.
+ */
+'use strict';
```
JaKXz: I don't see this in some other files, we should probably be consistent.

coreyfarrell: Added to the production sources of istanbul-lib-source-maps where it was missing.

coreyfarrell:
@JaKXz do you have any further comments on this?

JaKXz:
Sorry for the delay!

PR #306, opened 2019-02-21
feat: set medium colour to yellow

whitneyit:
As discussed in https://github.com/gotwarlost/istanbul/issues/542, this
PR brings back the yellow colour for medium.

There are many good points raised in that PR, and this PR, aims to
replicate address those concerns for this repo.

### DIFF
--- a/packages/istanbul-reports/lib/html/assets/base.css
+++ b/packages/istanbul-reports/lib/html/assets/base.css
@@ -172,16 +172,17 @@ table.coverage td span.cline-any {
 /* dark green */
 .status-line.high, .high .cover-fill { background:rgb(77,146,33) }
 .high .chart { border:1px solid rgb(77,146,33) }
-
-.medium .chart { border:1px solid #666; }
-.medium .cover-fill { background: #666; }
+/* dark yellow (gold) */
+.status-line.medium, .medium .cover-fill { background: #f9cd0b; }
+.medium .chart { border:1px solid #f9cd0b; }
+/* light yellow */
+.medium { background: #fff4c2; }
 
 .cstat-skip { background: #ddd; color: #111; }
 .fstat-skip { background: #ddd; color: #111 !important; }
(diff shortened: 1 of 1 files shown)


packages/istanbul-reports/lib/html/assets/base.css
```diff
@@ -172,16 +172,17 @@ table.coverage td span.cline-any {
 /* dark green */
 .status-line.high, .high .cover-fill { background:rgb(77,146,33) }
 .high .chart { border:1px solid rgb(77,146,33) }
-
-.medium .chart { border:1px solid #666; }
-.medium .cover-fill { background: #666; }
+/* dark yellow (gold) */
+.status-line.medium, .medium .cover-fill { background: #f9cd0b; }
```
coreyfarrell: This seems to me like a 'bright yellow' rather than dark yellow.  How would you feel about `#f1c50b` instead (for `.medium .chart` as well)?

whitneyit:
I'm fine with any _yellow_ colour. The colours I pulled were from the origin PR.

https://github.com/gotwarlost/istanbul/pull/597/files

coreyfarrell:
I just tried this out (nice utility btw).  I think this PR is an improvement for deuteranopia (listed as common).  Deuteranopia seems to make the current background red (for low coverage) appear as grey (the same as middle coverage).  Unfortunately green (high coverage) appears yellowish, but it appears as a different yellow than what we use for middle coverage.

New color scheme:
image

With deuteranopia:
image

Original with deuteranopia:
image

So I think this is an improvement to switch medium coverage away from grey.  I think it's less harmful for medium coverage to nearly look like high coverage than it is for low coverage to look exactly like medium coverage.

JaKXz:
Thanks @coreyfarrell! I am good to merge if you are

coreyfarrell:
Thanks for the contribution @whitneyit!

whitneyit:
This was good news to wake up to! Cheers all

PR #307, opened 2019-02-22
chore: ES2015ify test-exclude.

coreyfarrell:
* Use node.js 6 features
* Expand test coverage to 100%

---

Next step will be to add support for `getAllFiles` so we can centralize some more of the include/exclude logic here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,15 @@ node_js:
   - 10
   - 8
   - 6
+matrix:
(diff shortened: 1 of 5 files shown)


coreyfarrell:
A new test I added spawns a child_process, failing on Windows latest node only.  Possibly the same error we had on nyc?

coreyfarrell:
Looks like we're getting bitten by https://github.com/nodejs/node/issues/25484.  I'm going to push an additional commit to filter node.js 11 out from the Windows testing.

PR #309, opened 2019-02-23
feat: Add TestExclude.globSync to find all files

coreyfarrell:
I'm unsure if `extensions` should be an argument to `TestExclude.globSync` or if it should be added as an option to the object constructor.  If we go with the constructor option then should that have an effect on `TestExclude.shouldInstrument`?

This function takes a `cwd` argument because `nyc instrument <input> [output]` also takes the `--cwd` flag, although the `cwd` option is not looked at by this subcommand.  Maybe we can just pass `info` as `cwd` to the `NYC` constructor so it creates `TestExclude` with the correct `cwd`, then we could remove the `cwd` argument from this new function.

I've named this new function `globSync` with the plan that we might introduce an async version once we drop node.js 6.

CC @AndrewFinlay


AndrewFinlay:
@coreyfarrell I like the idea of `extensions` and `cwd` being constructor arguments, but I don't know enough about the use cases to make a definitive call on it.  

I was thinking about what the nyc walker is trying to achieve though, essentially we're trying to create a pre-filter here to limit the amount of work visitor has to do.  Currently we're filtering the `exclude`s, soon to be restoring the !`exclude`s , but we're not doing anything with the `include` list.  Effectively our current include filter is our file extensions list, but I figure that somehow that could be expanded to use the `include` list in the config.

coreyfarrell:
@AndrewFinlay This is a good point about the `include` list, this probably should be using the `include` list to filter.  I'm not sure if it's best to filter by extensions then by includes or to filter by includes then extensions.

I think this could be the source of an NYC bug (I don't have time to look it up right now).  I'll work on this more tomorrow.

AndrewFinlay:
@coreyfarrell I don't think the order matters too much, essentially you're finding the intersection of two sets so the result should be the same either way.  The only difference I can think of is that one order may take longer than the other.  I had a look at nyc to see what it does, but it looks like the order changes depending on which command you're using.

coreyfarrell:
@JaKXz some additional changes were made since you reviewed, is this still OK with you?

JaKXz:
:shipit:

PR #320, opened 2019-03-05
chore: Add test for negated exclude of file that is not included.

coreyfarrell:
Added test to ensure behavior doesn't accidentally change in the future.


JaKXz:
To be honest I'm surprised we don't have this test already but :shipit:

PR #325, opened 2019-03-06
chore(eslint): add rules

coreyfarrell:


coreyfarrell:
Not sure what the deal with node.js 11 on travis-ci is.  No failure when I tested locally.  Tomorrow I'll try resetting the branch and push one commit at a time to see which causes travis to fail.

coreyfarrell:
1. I tried `eslint-plugin-node`, it didn't find anything valid and had a bunch of false positives due to the way we have our lerna repo setup.  Specifically it complained about every instance of `require('chai')` as an unresolved import (because chai is a devDependency in the root package.json).  I'm not against looking at it again at some later time but for now I'd rather just stick with eslint-plugin-import.
2. I just tried that, unfortunately looks it did not fix the issue (node.js 11 has already failed on OSX).  Will have to explore further tomorrow.

coreyfarrell:
@JaKXz Looks like nothing was wrong with my change, we're dealing with a change of behavior in node.js 11.11.0.  I'm about to post a PR to display this so I can post a bug against node.js

PR #326, opened 2019-03-08
Modify .travis.yml to display regression in node.js 11.11.0.

coreyfarrell:


coreyfarrell:
I've determined that https://github.com/nodejs/node/commit/b338edbb0a is the cause of this failure.

PR #332, opened 2019-03-09
feat: Create @istanbul/nyc-config-hook-run-in-this-context.

coreyfarrell:
This base configuration helps enable `hook-run-in-this-context` in a way
that instruments files loaded via `require()` only once regardless of
node.js version.


coreyfarrell:
I'm open to other suggestions for how to accomplish this.  Also open to suggestions for a shorter name for this config module.

coreyfarrell:
> This is cool - should ~[some of]~ the other provided configs extend this to leverage the hook-require correctly?

That would be a breaking change for those other configs.  Currently the other configs do not enable `hook-run-in-this-context`.  I'm not familiar enough with typescript to know if enabling `hook-run-in-new-context` would be good or bad.  If `hook-run-in-this-context` is not enabled then `hook-require` needs to be enabled in all versions of node.js.

I believe it's pointless for nyc-config-babel to inherit this as it sets `instrument: false`, so I think all of the hooks are actually noop's.  In reality I think adding `hook-require: false` to nyc-config-babel wouldn't cause harm and would slightly optimize nyc for that config (actual useful hooking is done by `@babel/register` in that setup).

bcoe:
@coreyfarrell does this put you in a position to fix the tests that are failing in `11.11`?

coreyfarrell:
I already fixed the test that was failing in 11.11.0, see 8d66d2539e55a46d13fc658f6b62f7a188c12535.  That test is not using this new module as the test does not use nyc, it's an istanbul-api test.

PR #333, opened 2019-03-09
feat: Add common *.config.js dev configs to default exclude.

coreyfarrell:
Ref istanbuljs/nyc#1018

### DIFF
--- a/packages/test-exclude/index.js
+++ b/packages/test-exclude/index.js
@@ -172,6 +172,8 @@ function getExtensionPattern(extension) {
 
 const exportFunc = opts => new TestExclude(opts);
 
+const devConfigs = ['ava', 'babel', 'jest', 'nyc', 'rollup', 'webpack'];
+
 exportFunc.defaultExclude = [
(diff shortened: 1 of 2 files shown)


packages/test-exclude/index.js
```diff
@@ -180,7 +180,7 @@ exportFunc.defaultExclude = [
     '**/*{.,-}test.js',
     '**/__tests__/**',
     '**/node_modules/**',
-    '**/babel.config.js'
+    '**/{babel,nyc}.config.js'
```
JaKXz: you could also add `jest`

coreyfarrell: Sure.  Going to also add `ava`.  Any others you think?

JaKXz: Not off the top of my head

JaKXz: `webpack`?

coreyfarrell: Added `rollup` as well.

PR #343, opened 2019-03-30
fix copy of arrow png on html report

lukeapage:
This meant that the sort indicator arrows were broken (on windows 10 chrome anyway)


coreyfarrell:
@lukeapage good catch, thanks!

Issue #69, opened 2017-06-14
New design for HTML report


J-Sek:
Some time ago I created ["hacky" way](https://github.com/J-Sek/istanbul-coverage-report-extensions) of getting some UX improvements to HTML Istanbul report. I would love to make these improvements available to all developers, so I plan to port them (or just a part we agree on) to Istanbul source.

Would be great if you could let me know if you like this idea or not, so I can proceed with preparing clean implementation. :)

Please check [live demo](http://istanbul-report-demo.surge.sh) I managed to generate for Istanbul sources.

Problems with current design:

- hard to read, flat list, noisy design
- navigation between static pages, loosing context

New design:

- tree structure
- opening file coverage in modal dialog
- summary chart with meaningful info/warning

Known issues/limitations of current implementation:

- iziModal behavior when displaying report in iframe
- missing routing (cannot share link to an open modal)
- report should rather be served on the server in order to fetch all pages

bcoe:
@J-Sek wow, this is pretty darn slick! I'm a bit slammed this week with work at npm -- but am really digging this.

I'm definitely open to redesign, curious to hear other's thoughts (CC: @gotwarlost, @isaacs).

isaacs:
The modal iframe doesn't seem to work so well on small iOS screens. I'll take a look at it on a bigger device in the next few days.

isaacs:
Ok, looked on a larger screen.

There are some aspects of this I like, and some I definitely don't.

Like:

1. The full tree is nice.  Clicking through separate pages per-folder is kind of tedious.
2. I can see how sorting files by coverage is super handy.

Don't like:

1. WAY too much motion.  It's distracting for me, and probably will cause nausea or dizziness in anyone with certain vision and/or vestibular issues.
2. The model feels pretty klunky.  I'd actually be ok with a frameset, but having two scrollable windows overlaid on one another such that a single scroll action can cause unpredictable effects is very weird to me.
3. When sorting by anything other than filename, the folder structure is completely lost.
4. I don't like saying "there's a x% chance that you won't notice broken logic".  I frequently write bugs in 100% test covered code.  There is a 100% chance that a program has broken logic.
5. Sometimes sorting by coverage levels makes the whole file list disappear:

    ```
    TypeError: Cannot read property 'options' of undefined
        at iziModal.min.js:1
        at dispatch (jquery-3.1.0.min.js:3)
        at q.handle (jquery-3.1.0.min.js:3)
[...]

Issue #273, opened 2019-01-15
proposal: change HTML report to better standardize across Babel and V8


bcoe:
I would like to make two changes to to HTML reports, motivated by the implementation of test coverage in V8, moving us towards a common ground for reporting that makes sense for both V8 and Istanbul's babel-based instrumentation.

My motivation is the fact that, over time, I would like to gradually migrate major code-bases like Node.js over to using Chromium's built in coverage (see: https://github.com/nodejs/node/pull/25157); while adopting Istanbul's output format and reports format as a standard.

Here is what I'm proposing:

## 1. rename `branch` coverage to `block` coverage in reports

V8's coverage has no concept of branches, only blocks. The following code represents three blocks:

```js
{
  const a = 'hello world'
}
const b = c ? 'hello' : 'world'
```

The significant difference between Istanbul and V8 is that Istanbul will not include blocks that have only one logical path (only tracking logic that branches). 

However, in both cases, moving towards 100% block coverage will ensure that a test has been written for every conceivable branch -- this is why I think moving to V8's `block` concept is a reasonable compromise.

## 2. retire the `statement` column, in favor of `lines`.

v8 only has the concept of blocks and functions when collecting coverage; blocks can either span a partial line, a full line, or multiple lines.

Istanbul's Babel instrumentation tracks statements (which are more limited in their scope than blocks):

```
{
  const foo = bar;
}
```

☝️the above code is recorded by V8 as a single block, but in Istanbul as multiple statements (`AssignmentPattern`, `BlockStatement`).

... surfacing this in reports feels like a leaky abstraction.

Developers think more in terms of lines of code covered (statements are an implementation detail). Let's show `lines` in reports, which works for both V8 and Istanbul.

----

I'm excited that a few small changes to Istanbul reports will but us in a better position to support test coverage across both the V8 and Babel implementations.

CC: @demurgos

coreyfarrell:
Any thoughts about dropping collection of statements data entirely?  This could simplify the instrumentation code and decrease performance penalty of running instrumented code.

I don't think this change should be made to HTML reports only, if we proceed I think we should remove the column from the text report (and any others) as well.

Take the following examples:
```js
const b = c ? 'hello' : 'world'
const d = c || 'alternative'
```

What does V8 coverage say about this?  Can we determine that true and false are hit for both lines?

bcoe:
[...]

PR #345, opened 2019-04-01
WIP: Modern html report

lukeapage:
I'm implementing something like https://github.com/istanbuljs/istanbuljs/issues/69 but I won't be taking that implementation as there's a few things I think can be improved from it (and its a hack on top of the current report - this new report aims to be a new thing).

Preview: http://istanbul-modern-report.surge.sh

Things left to do:

 - [x] routing is only partially working - you can't press back to navigate through changes. I'm not sure tbh if its better with or without though?
 - [x] file coverage pages link to non-existent index files
 - [x] work out why its hanging on the linter during the test task.. 
 - [x] add some tests (not sure though if I should go so far as to add front end tests.. potentially using a different framework from the node tests ?!)
 - [x] review - I have a few questions particularly around integration with this code base
 - [x] save the state of expanded lines
 - [x] consider showing uncovered instead of covered

from code review:

 - [x] add the feature to select what metrics to show and default to removing statements.
 - [x] Set cursor in CSS where appropriate (expanding folders for example).
 - [x] Default sort is by filename, it would be nice if the sort icons were 3-state. So clicking multiple times on the branch coverage would cycle between:
    ascending by branch
    descending by branch
    ascending by filename (do not sort by branch).
 - [x] Just to point out currently the sort goes to descending first, so if you click on lines % the files with 100% coverage show first. I think ascending first would be better, people want to see where coverage is missing? I know for the existing report clicking the sort goes to descending first, I think this is a bug.
 - [x] I wonder if Nested would be the best default view?
 - [x] Right now clicking the + or clicking istanbul-api link do the same thing. Is it possible/planned to make it so clicking the word istanbul-api actually opens that folder (IE hides everything else and adds a link to go to parent / root)? I think this could help reduce information overload.
 - [x] change licence in modern-html to MIT
 - [x] work out the changes to istanbul-api and if its okay or should be discussed seperately


packages/istanbul-report-modern-html/LICENSE
```diff
@@ -0,0 +1,24 @@
+Copyright 2012-2015 Yahoo! Inc.
```
lukeapage: ?

packages/istanbul-api/lib/reporter.js
```diff
@@ -88,6 +97,32 @@ Reporter.prototype = {
                 )
         );
 
+        // TODO:
```
lukeapage: nyc has its own implementation too.. and that doesn't allow configuration
so also.. change both?

coreyfarrell:
The idea of reworking the current HTML report has been on my mind.  Honestly I had not considered using react.  I know it's very popular and JavaScript template literals are a newer feature but my preference is to use standard modern JS syntax.

I had planned to rewrite handlebars templates to use JS template literals to be processed with [viperhtml](https://github.com/WebReflection/viperHTML/).  I'd be open to other options, my choice of viperhtml is purely due to familiarity with hyperHTML (same author, same API).  Honestly this idea has stalled, not due to any technical reason just lack of time on my part.

Please do not take this comment as discouragement or even as any kind of official position of the istanbuljs project, I'm just sharing my personal thoughts.

lukeapage:
@coreyfarrell Thanks for some feedback.
I planned to use react - like you, its because it is what I am most familiar with.
I originally thought about a completely vanilla js solution.. but thinking about having expanding/collapsing, sorting, configurable columns, choice between nested/package/flat inside the report etc. I felt it could potentially be a bit flashy if the whole dom was just replaced each time.
React gives you a) templates b) a very simple state management system
So I thought I would be able to get the existing functionality working pretty quickley

coreyfarrell:
So is the intent to do browser-side rendering / updates or will static renderings be generated to ./coverage/ then updated inside the browser?

lukeapage:
My intention was only to work on summary page and make it a single page that is fully rendered browserside. It would keep the existing report essentially as static html.
If this was a web site I would render the html start state into the html and hydrate it to get the best of both worlds - but it doesn’t seem necessary for this usecase.

bcoe:
@lukeapage :wave: don't have too much time to give feedback yet, but thanks a ton for diving in and taking this on, it would be lovely to dust off the design a bit.

lukeapage:
@bcoe @coreyfarrell 
I've published an example of what I have so far: http://istanbul-modern-report.surge.sh

I'm no design wizard.. its very much like the existing report but with changes I'd like to see for improved navigation and drilling down of data. I'd welcome feedback.

If your not happy with taking this, I always have the option of publishing as a seperate module (either on my own or within istnbuljs org).. the downside is less visibility and I guess you wouldn't want specific code added to nyc to support it. but it would work for my own use case where we currently create a report directly.

Things left to do:

[ ] - routing is only partially working - you can't press back to navigate through changes. I'm not sure tbh if its better with or without though?
[ ] - work out why its hanging on the linter during the test task.. 
[ ] - add some tests (not sure though if I should go so far as to add front end tests.. potentially using a different framework from the node tests ?!)
[ ] - possibly add the feature to select what metrics to show (or do in a later pr).
[ ] - review - I have a few questions particularly around integration with this code base

packages/istanbul-report-modern-html/LICENSE
bcoe: that's a blast from the past isn't it :stuck_out_tongue: the original Istanbul library grew out of work in Yahoo; If you're doing a rewrite of the HTML report, let's license as MIT or ISC, to "Contributors".

bcoe:
@lukeapage I'll carve out some time to provide more of a design review next week. At first glance:

* I'm very excited about being able to delve in to sub-folders with a click.
* I like that the sorting functionality is more explicit, and I think it's neat that you can choose to hide files with low/medium/high coverage.
* I think the amount of information on the screen is a bit dense; this is partially hold over from the old report breaking out `branches`, `functions`, `statements`, and `lines` ... here's a thought:
  * as discussed [here](https://github.com/istanbuljs/istanbuljs/issues/273) I think having both `lines` and `statements` is a bit of overkill (V8 doesn't really have the concept of statements).

coreyfarrell:
So far I've only reviewed the demo page, I haven't had a chance to look at actual code.  I agree with what @bcoe has said the listing page is a bit of information overload.  This is really nice though.  I especially like that this does not pull heavy dependencies into production (like handlebars for the current HTML report).

Some comments (in random order, not all high priority):
* Set `cursor` in CSS where appropriate (expanding folders for example).
* Default sort is by filename, it would be nice if the sort icons were 3-state.  So clicking multiple times on the branch coverage would cycle between:
   1. ascending by branch
   2. descending by branch
   3. ascending by filename (do not sort by branch).
* Just to point out currently the sort goes to descending first, so if you click on lines % the files with 100% coverage show first.  I think ascending first would be better, people want to see where coverage is missing?  I know for the existing report clicking the sort goes to descending first, I think this is a bug.
* I wonder if `Nested` would be the best default view?
* Right now clicking the `+` or clicking `istanbul-api` link do the same thing.  Is it possible/planned to make it so clicking the word `istanbul-api` actually opens that folder (IE hides everything else and adds a link to go to parent / root)?  I think this could help reduce information overload.

lukeapage:
@bcoe @coreyfarrell I've updated http://istanbul-modern-report.surge.sh with most of the comments and am tracking things to be done as checkmarks in the main description.

.eslintrc
```diff
@@ -1,42 +1,66 @@
 {
-  "extends": [
```
lukeapage: Big changes because prettier formatted this file.

packages/istanbul-api/lib/reporter.js
```diff
@@ -50,7 +50,13 @@ Reporter.prototype = {
             if (this.config.verbose) {
                 console.error('Create report', fmt, ' with', rptConfig);
             }
-            this.reports[fmt] = libReports.create(fmt, rptConfig);
+            let report;
```
lukeapage: I need to make the same changes to nyc. The question is if you want something bigger done.

packages/istanbul-api/test/reporter.test.js
```diff
@@ -35,6 +35,8 @@ describe('Reporter', () => {
 
                 const reporter = new Reporter(config, opts);
 
+                reporter.reports = { test: {} };
```
lukeapage: can be removed, doesn't hurt but currently not needed (was a previous refactor)

packages/istanbul-report-modern-html/lib/index.js
```diff
@@ -0,0 +1,154 @@
...
+            coverageSummary.functions,
+            'functions',
+            context
+        ),
+        lines: this.getMetric(coverageSummary.lines, 'lines', context)
+    };
+
+    return {
+        file: node.getRelativeName(),
+        output: parent && this.linkMapper.relativePath(parent, node),
```
lukeapage: not currently used. either use this or remove it

packages/istanbul-report-modern-html/assets/modern.css
```diff
@@ -0,0 +1,347 @@
...
+.pad1 {
+    padding: 10px;
+}
+.space-left2 {
+    padding-left: 55px;
+}
+.space-right2 {
+    padding-right: 20px;
+}
+.center {
```
lukeapage: can be further cleaned up

packages/istanbul-report-modern-html/src/index.js
```diff
@@ -0,0 +1,119 @@
...
+                                    {...child}
+                                    key={child.file}
+                                    metricsToShow={metricsToShow}
+                                />
+                            ))}
+                        </tbody>
+                    </table>
+                </div>
+                <div className="push" />
+                {/* for sticky footer */}
```
lukeapage: comment was in old report. seems invalid?

coreyfarrell: See http://istanbul-modern-report.surge.sh/istanbul-api/lib/input-error.js.html - `Code coverage generated by istanbul at Sun Apr 14 2019 21:32:36 GMT+0200 (Romance Summer Time)` is pinned to the bottom of the page as long as the page is not scrolling.  Unfortunately the same does not happen for the directory view.

lukeapage:
Would uncovered lines be a better column than covered? Since you mostly want to use the report to find things not tested..?

packages/istanbul-report-modern-html/src/index.js
lukeapage: Fixed

packages/istanbul-report-modern-html/src/filterButtons.js
```diff
@@ -0,0 +1,48 @@
+import * as React from 'react';
```
JaKXz: maybe my knowledge is out of date, but can't we just `import React from 'react'`? The `React` object just needs to be in the namespace, to me this implies we're going to do something with it... again I could be wrong though.

packages/istanbul-report-modern-html/src/routing.js
```diff
@@ -0,0 +1,47 @@
...
+        encodeURIComponent(fileFilter),
+        expandedLines.map(encodeURIComponent).join(',')
+    ];
+    const newUrl = `#${params.join('/')}`;
+
+    if (newUrl === location.hash) {
+        return;
+    }
+
+    window.history[isReplace ? 'replaceState' : 'pushState'](null, '', newUrl);
```
JaKXz: this is interesting... can it not be done in a routing library? I mean I'm totally fine with this but it could be cleaner, no?

packages/istanbul-report-modern-html/assets/modern.css
```diff
@@ -0,0 +1,257 @@
...
+a:hover {
+    text-decoration: underline;
+}
+.strong {
+    font-weight: bold;
+}
+.space-top1 {
+    padding: 10px 0 0 0;
+}
+.pad1y {
```
JaKXz: This suggestion is probably overkill, but, I like having a consistent framework for things like this (not that you have made inconsistencies here!) -- I'm a huge fan of https://styled-system.com/ 

What do you think?

packages/istanbul-report-modern-html/lib/index.js
```diff
@@ -0,0 +1,181 @@
...
+        const sourcePath = path.dirname(this.getPath(source));
+        return path.relative(sourcePath, targetPath);
+    },
+
+    assetPath(node, name) {
+        return this.relativePath(this.getPath(node), name);
+    }
+};
+
+function ModernHtmlReport(opts) {
```
JaKXz: Let's just use a `class` here?

packages/istanbul-report-modern-html/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "description": "istanbul report - modern html",
+  "author": "Luke Page",
+  "main": "index.js",
+  "files": [
+    "index.js",
+    "lib",
+    "assets"
+  ],
+  "scripts": {
+    "test-disabled": "mocha --recursive",
```
JaKXz: This should be made compatible with the other packages/* test scripts so coverage is collected :)

packages/istanbul-report-modern-html/src/summaryTableLine.js
```diff
@@ -0,0 +1,188 @@
...
+            ? [
+                  ...expandedLines.slice(0, expandedIndex),
+                  ...expandedLines.slice(expandedIndex + 1)
+              ]
+            : [...expandedLines, prefix + file];
+
+        return (
+            <>
+                <a
+                    href="javascript:void(0)"
```
JaKXz: Hm... Is this necessary? I would rather use an unstyled (css-reset) button to indicate actions. ideally `<a>`s are for navigation right?

packages/istanbul-report-modern-html/src/filterButtons.js
lukeapage: I used what I am used to.. and it works fine (flow requires "React" to be imported as * as).  but what you've suggested is what the official doc examples look like, so I'll change it to that

packages/istanbul-report-modern-html/src/routing.js
lukeapage: I looked at a couple of routing libraries - I wanted to use it, but for this simple case, it was more effort than implementing this

packages/istanbul-report-modern-html/assets/modern.css
lukeapage: This css file is inherited from the existing report - I would have tried styled components if I was starting from scratch, but actually, I took the existing css, removed unused classes and added a small amount of css at the end.
Someone would be welcome to refactor this and improve upon the styling in a later PR imo.

packages/istanbul-report-modern-html/lib/index.js
lukeapage: This is just following the existing pattern in other modules - so its up to the maintainers..

packages/istanbul-report-modern-html/package.json
lukeapage: yep, this is the last tick mark - add tests. If I leave as is without tests then it causes the build to fail

packages/istanbul-report-modern-html/src/summaryTableLine.js
lukeapage: I changed the expand button to button. The other link I've kept `<a>` for consistency with the file links (and it does change the page and url anyway)

packages/istanbul-report-modern-html/lib/index.js
coreyfarrell: The pattern is due to many of the modules pre-dating node.js 4.  We now require node.js 6 (soon to be node.js 8), so we've been working towards modernizing the code.  It's a process and we haven't gotten around to updating any of the existing reports.

coreyfarrell:
@lukeapage thanks for your great work.  One thing I'd like to see is for the code to be moved into the istanbul-reports package, so `packages/istanbul-reports/lib/modern-html/`.  This would make it unnecessary to modify other packages to recognize that the new report exists.

@bcoe @JaKXz Do we want to plan to replace the current report in a future semver-major?  @lukeapage what kind of browser support do we expect from this?  I assume JavaScript is mandatory or will/can it gracefully degrade?

lukeapage:
@coreyfarrell
Because it requires multiple summarizers even moving it into the reports package will not make it invisibly available to other packages. If that wasn’t a problem I could have made the reports package require in the modern html report and still have it invisible. Given that I prefer it to be separate- it simplifies the setup of that package without polluting all reports. If you have some ideas of a bigger refactoring to let reports choose or consume multiple summarizers, then let me know what you think.

Re browser support I would expect ie10+ - React is pretty good and it is transpiled. I haven’t tested it though.

Without js the page shows nothing..

I would expect for that reason there is a reason to keep the old report around instead of replacing it entirely.

lukeapage:
Perhaps make it available and then change it to a default in a future major version bump?

lukeapage:
It would make sense for a single api to do any report - that could be through the api package and I could code that so that consumers of it just saw it as another report.. but nyc doesn’t use it, it uses packages directly.. and I saw a comment talking about deprecating the api package - so where can the logic that combines reports and summarizers live? Or the intention was that it should be the user of Istanbul to implement that?
Hopefully you understand my problem now...

bcoe:
@lukeapage @coreyfarrell given the prevalence of the existing HTML report (as an example it's what's used for Node's nightly [coverage generation](https://coverage.nodejs.org/coverage-7167eb2f12a5070c/index.html), I'd like to hold our horses a bit on planning to replace the existing HTML report with this one. I have few things on my mind I think we should address first:

1. how big is the new HTML report vs., the legacy one?:
  * I see that we use `rollup` to bundle everything, so if I understand correctly, someone installing `istanbul-reports` wouldn't need to install any dependencies? (this is great 😄)
1. I don't love the name "modern-html-report", would "interactive-html-report", or "single-page-html" be a better name?
1. I still think there's some information overload on the screen; I'd like to potentially get a designer friend to take a look, mainly:
  * the "coverage:" picker could maybe just be a check-box that lets you hide anything that's 100% function, line, and branch coverage.
  * for "Summarizer:" I think we should just pick the view we like best; I'm pretty happy with nested, don't know about other folks?
  * I feel like the `covered` and `total` columns could be combined like in the old report, which would help cut down on information overload.

In general, really happy with the direction this is going in 👍 having fun walking through the file tree without reloading content 😄

lukeapage:
Of course, no hurry, I was just suggesting a possible eventual path.

1. I will get some real numbers, but
a. Yes it’s done via dev dependencies, so npm install Istanbul reports does *not* add Rollup/react etc
b. The js is much bigger (approx 140kb), the combined html is smaller and less files (Istanbul repo is 70kb vs 154kb) so this adds 50kb to this repo. With more files and folders the difference will be less, eventually becoming smaller
2. Sure yes, that would be great, I am no designer, but there is some reasoning behind things
a. Imo just hiding 100% coverage is not useful unless your a project that aims for 100% coverage. At my work we have several thousand files and not many are 100%. I want to see files with low coverage.. in a range.. Because maybe I want to see all lows, maybe all mediums if I don’t have any lows..
In a perfect world I wanted some kind of filtering with a nice ux .. but I dont have the time or expertise to do this
b. I prefer nested too - it was a surprise when I started this that it was configurable - but actually isn’t configurable in nyc. Even in the api, you can’t specify different summarizers for different reports.
However I do think flat is useful - in fact it was my 2nd biggest feature request - id like to be able to look at my project and answer questions like - what’s the most complex code with low coverage - flat enables that combined with filtering and sorting.
I’d be happy just having nested and flat, and this would mean that maybe it could have the same api as other reports.. but.. it would rely on the chosen summarizer, which defaults to package.. no one has yet engaged me on answering my api/code questions about how to sort this out..
c. I’d like to be able to sort by files with the most uncovered.. but we have covered and total.. which I think provides enough functionality, but isn’t perfect as you say, because it’s cluttering.
One option would be to show the covered lines/total but sort by uncovered.. or just show uncovered? I think that’s what I might do if it was me.
If you combine them back how it was though, I think you lose too much in terms of ability to dig into the data.. which was my main motivation.. I have a very large repo and the old report makes it very hard to get answers to questions.

Also I am running out of time that I can spend on this - I will be able to finish things off and make small changes, but if you need something big, I will have to leave it - for my purposes I can put this in my own repo and use it, but of course I would rather it’s jmproved and I’d rather it was in core.

lukeapage:
@bcoe also.. re: naming. Please decide a new name - I called it modern report because you suggested that here: https://github.com/istanbuljs/istanbuljs/issues/69#issuecomment-317205362

bcoe:
@lukeapage I very much appreciate the work you've put into this, and thanks for clarifying your motivation for the `coverage` and `summarizer` widgets -- I think maybe with some playing we can keep this functionality, but just make it more obvious to the laymen what the controls are doing...

## A couple UX thoughts

Maybe just adding a `?` widget that links to docs or something eventually; explaining `low`/`medium`/`high` watermarks.

As for the `Summarizer`, maybe we could compromise and just allow the user to toggle between Nested and Flat?

## Naming

I vote we go for the name `html-spa`:

* it's short, so it's not too much of a hassle to type in configuration.
* a single-page-app is the main difference this introduces over the html report.

## In General

If @coreyfarrell doesn't object, I'd say let's land this and iterate; I'm fine with polish happening post-hoc.

If I understand, we're adding a about 50kb in size to the repo, but getting a fancy-pants new interactive report?

---

Very much appreciate the work you've put in @lukeapage, and I think this will provide a great foundation for folks to iterate on.

I also do like the idea, once we've iterated for a bit, that we'd retire the old `html` report, thinking this repo size a bit, and making your reporter the canonical one.

coreyfarrell:
@lukeapage your other PR's have been merged, this PR now conflicts because of that.  Would you have a moment to just take care of those conflicts?

Once that is done I would like to see this new code live in `packages/istanbul-reports/lib/html-spa/` instead of `packages/istanbul-report-modern-html/`.  This would eliminate the need to modify `istanbul-api` and would make this new report available to existing users of istanbul reporting system (all they would have to do is reference `html-spa`).  I can take care of moving everything into istanbul-reports if you don't have time but I would like to ask that the `istanbul-api` changes be excluded from this PR.  They will become unnecessary once this new report is moved into `istanbul-reports` - istanbul-api, nyc and any other users will just automatically find `html-spa` by name.

@bcoe I'm good with merging this minus the istanbul-api changes but I'm leaning towards lets get #375 shipped first, I'd like to get that out with priority.

lukeapage:
@coreyfarrell it will require the same changes to api and nyc if the package is moved as I explained here : https://github.com/istanbuljs/istanbuljs/pull/345#issuecomment-485230234

You need to engage me on fixing the summarizer problem if you want it transparent.

To be clear : once I make flat a toggle, it only requires one summarizer.. but a. The default is package and in nyc it is not configurable. I see how package makes sense for some reports, but not the html one.

I will make the other changes, hopefully tonight.

lukeapage:
Also.. the index file conflicts with the existing html report.. so I guess by default all reports shouldn’t include the new one? Only if specified?

lukeapage:
What I could do is have this as you suggested, no other changes.. and then make a future Pr to sort out summarizers, with changes to the api and nyc packages.. but I will still need help on what direction you would like or would accept.

lukeapage:
Okay, I refactored into `instanbul-reports` - I have left a small change in `istanbul-api` and if you are okay with it I will make a PR to nyc to do the same.

Am in the middle of adding some tests and I still have the tooltip on the watermarks filtering to do.

packages/istanbul-api/lib/reporter.js
```diff
@@ -91,7 +92,12 @@ Reporter.prototype = {
         const tree = this.summarizer(coverageMap);
         Object.keys(this.reports).forEach(name => {
             const report = this.reports[name];
-            tree.visit(report, context);
+            if (name === 'html-spa') {
+                const nestedTree = libReport.summarizers.nested(coverageMap);
```
bcoe: this seems reasonable to me, we always process the `nested` list, and just flatten it if you're in the flat mode?

bcoe:
@lukeapage don't feel you need to block on the tooltip, it would certainly be cool if we launch with it -- but doesn't need to block you.

lukeapage:
@bcoe yep. Feel free to merge this once you are happy. I’ve just realized now there are package Json conflicts I can resolve tomorrow. I will add tooltip either to this Pr or a new one of this is merged before I get to it.

bcoe:
@lukeapage, @coreyfarrell was digging through the code, and couldn't figure out why you needed to use:

`const nestedTree = libReport.summarizers.nested(coverageMap);`

Rather than the approach used by the other reporters -- given that I assume you rework this data any ways when you're switching between a flat and nested report?

How hard of a refactor would it be to eliminate this API change -- I can understand's Corey's point of view.

Maybe as a first step in the `html-spa` we could just build the nested report, before running all the same logic?

---

@coreyfarrell does this have to block mering, wonder if we could land this as the base, and then refactor to eliminate the API change before we release the module.

lukeapage:
> Rather than the approach used by the other reporters -- given that I assume you rework this data any ways when you're switching between a flat and nested report?
> 
> How hard of a refactor would it be to eliminate this API change -- I can understand's Corey's point of view.
> 
> Maybe as a first step in the `html-spa` we could just build the nested report, before running all the same logic?

@bcoe @coreyfarrell 

I do a transform from nested to flat - and thats easy, its 10 lines of code and it just drops the directories.

I don't have code to turn flat into nested.. There are 3 ways I can do this:

1. make `html-spa` work no matter what summarizer is used. Basically i'd full out the leaf nodes so it was always flat, then I'd re-implement the logic inside the nested summarizer in order to make the flat nodes into a nested view and then re-visit myself. 
downsides: code duplication, lots of code, hacky
upsides: no change to the report api

2. make `html-spa` follow the existing reports.. if packages summarizer is used (which in nyc is the default and *cannot be changed*) then you get packages and flat. if you use nested, you get what I intended. If flat is used, it looks wrong because you get flat and flat.
downsides: user can make bugs if they don't make sure to create the report using the right summarizer
upsides: no changes anywhere outside `html-spa`

3. as per (2) but change istanbul-api and nyc to use nested when its `html-spa`.
downsides: 3 line code change in istanbul-api (which you want to deprecate without users having an alternative)
upsides: all works fine

I went with (3). I can roll back to (2) so there are no changes to the api with the specified downsides.
You can also discuss whether I should make nyc use nested just for `html-spa` or continue to use package for everything.

What I would do to sort this out:
Create a neat tidy and straight forward report interface similiar to whats in istanbul-api but thats say 100 loc. put it in istanbul-reports, recommend that as an entry point, change nyc to use that. Then I'd make that code always use nested for `html-spa`.

lukeapage:
@coreyfarrell what do you want to do? If you want I can revert all changes to instanbul-api to get this merged - it will still work and be usable by people, it just won't have the benefits of a nested summarizer - then you can have time to decide what you want to do.

bcoe:
@lukeapage no huge blockers on my end, just don't have many cycles to land this this week; I'll try to carve out some time to see it over the finish line this weekend.

coreyfarrell:
@lukeapage I would prefer that we remove the `istanbul-api` changes.  I'd like to get this merged, then I think the next step will be to work on the API so that reports are given something resembling a summarizer factory, then each report can request the summarizer it needs.  Obviously changing the API of reporting will be semver-major but I think this is the right way forward (consistent API for all reports).

lukeapage:
@coreyfarrell thanks for the plan, I reverted api changes. I think its okay to merge now.

coreyfarrell:
@lukeapage thanks for working with us to create this new report.  Some further work is still likely but I feel that the best way forward is to get the bulk (new implementation) merged so any additional tweaks can be done as small isolated changes.

PR #348, opened 2019-04-02
chore: Add greenkeeper.json configuration.

coreyfarrell:
This (hopefully) configures greenkeeper to report updates for the actual
packages.  This will also enable testing of in-range upates.

The ignore list is modules that require node.js 8, we won't be dropping
node.js 6 until nyc@15.

### DIFF
--- a/greenkeeper.json
+++ b/greenkeeper.json
@@ -1,0 +1,15 @@
+{
+  "groups": {
+    "default": {
+      "packages": [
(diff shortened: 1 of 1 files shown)


coreyfarrell:
I had hoped that greenkeeper would complain if it didn't like the glob pattern, looks like it won't do that until greenkeeper.json is merged to master.

PR #351, opened 2019-04-02
fix: Remove `**/node_modules/**` from defaultExclude.

### LINKED ISSUE #347: test-exclude: Remove `**/node_modules/**` from the default exclude list.
We should remove this from the default list, just let `excludeNodeModules: true` (the default) add it.  This will make it easier for projects to disable excludeNodeModules (they won't have to override the default excludes to actually remove `**/node_modules/**`).

I'm on the fence about this being a breaking change or not.

coreyfarrell:
This is appended to the `exclude` list unless excludeNodeModules is
disabled.  Removing from the default list makes it easier to use
excludeNodeModules as you no longer need to reset the `exclude` array to
get coverage of node_modules.

Fixes #347

### DIFF
--- a/packages/test-exclude/index.js
+++ b/packages/test-exclude/index.js
@@ -57,7 +57,7 @@ class TestExclude {
             this.excludeNodeModules &&
             !this.exclude.includes('**/node_modules/**')
         ) {
-            this.exclude.push('**/node_modules/**');
+            this.exclude = this.exclude.concat('**/node_modules/**');
         }
 
         this.exclude = prepGlobPatterns([].concat(arrify(this.exclude)));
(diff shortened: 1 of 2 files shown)


packages/test-exclude/index.js
```diff
@@ -57,7 +57,7 @@ class TestExclude {
             this.excludeNodeModules &&
             !this.exclude.includes('**/node_modules/**')
         ) {
-            this.exclude.push('**/node_modules/**');
+            this.exclude = this.exclude.concat('**/node_modules/**');
```
coreyfarrell: This may appear unrelated but now `this.exclude === exportFunc.defaultExclude` which does not contain `**/node_modules/**` so without this shallow copy we would append the default list (this was caught by the tests).  Previously we were appending any `opts.exclude` that did not include `**/node_modules/**`, though that was unlikely to cause any real problem.

coreyfarrell:
CC @hershmire just so you know.

coreyfarrell:
Yes `excludeNodeModules` is true by default and is processed during the TestExclude constructor.

PR #352, opened 2019-04-02
fix: Be more friendly to ts-node.

### LINKED ISSUE #336: instrumenter: Be more friendly to ts-node/register
Per istanbuljs/nyc#1002 the instrumenter generates two pieces of code that are unfriendly to ts-node/register.  One is easy to fix, `coverageData.hash = hash;` should not be set separately but should be included in the declaration of `coverageData`.

The second issue with `new Function` is more difficult.  See b12a07e35918e5d17be66c5382e9a38b092fae32 for source of `var Function = function () {}.constructor;`.  I'm not sure if this can be fixed.

coreyfarrell:
ts-node gets angry when coverageData.hash is set after coverageData is
declared, add it to the initial object instead of setting it separately.

Fixes #336

---

Note for the instrumentation to be error-free with ts-node requires getting rid of the `var Function = (function(){}).constructor;`.  I know this was added for a reason, I don't want to create an explosion of new options.  So for now the `istanbul-lib-instrument` answer is to set the following options:
```js
{
  coverageGlobalScopeFunc: false,
  coverageGlobalScope: '(new Function("return this"))()'
}
```

I believe using these options would produce ts-node compatible code.  Unfortunately these options are not exposed by nyc or babel-plugin-istanbul.

### DIFF
--- a/packages/istanbul-lib-instrument/src/visitor.js
+++ b/packages/istanbul-lib-instrument/src/visitor.js
@@ -534,7 +534,6 @@ const coverageTemplate = template(`
         if (coverage[path] && coverage[path].hash === hash) {
             return coverage[path];
         }
-        coverageData.hash = hash;
         return coverage[path] = coverageData;
     })();
 `);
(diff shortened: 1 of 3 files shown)


coreyfarrell:
@addaleax I wanted to ping you on this since you added `var Function = function() {}.constructor;` in #131.  Do you see any problem with the second commit of this PR?

PR #356, opened 2019-04-05
feat: Add bigInt and importMeta to default parser plugins.

### LINKED ISSUE #338: istanbul-lib-instrument: Add `bigInt` to default babel parser plugins.
I'm not sure this should happen or if it should be considered a breaking change.  This is by request of @bcoe at https://github.com/istanbuljs/nyc/issues/947#issuecomment-474464888

coreyfarrell:
This allows code with these features to be instrumented without having
to reconfigure the parser plugins.

Fixes #338

### DIFF
--- a/packages/istanbul-lib-instrument/src/instrumenter.js
+++ b/packages/istanbul-lib-instrument/src/instrumenter.js
@@ -21,9 +21,18 @@ export function defaultOpts() {
         ignoreClassMethods: [],
         sourceMapUrlCallback: null,
         debug: false,
+        /* babel parser plugins are to be enabled when the feature is stage 3 and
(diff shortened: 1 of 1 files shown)


packages/istanbul-lib-instrument/src/instrumenter.js
```diff
@@ -21,9 +21,18 @@ export function defaultOpts() {
         ignoreClassMethods: [],
         sourceMapUrlCallback: null,
         debug: false,
+        /* babel parser plugins are to be enabled when the feature is stage 3 and
```
bcoe: love the addition of this comment, great stewardship.

PR #362, opened 2019-04-15
Fix handlebars security issue: https://npmjs.com/advisories/755

dprothero:


coreyfarrell:
The dependency `"handlebars": "^4.1.0"` will allow the non-vulnerable version to be installed so the issue is not here.  The issue is that nyc@13.3.0 bundles dependencies.  We're in the processes of releasing nyc 14.0.0 which will not include bundled dependencies and does not contain this vulnerability.

Worth noting this is a false positive as istanbul-reports does not use untrusted templates.

PR #364, opened 2019-04-15
Update all packages with npm audit fix to ensure known vulnerabilities

adamkingit:
Sorry my previous PR was a quick attempt that didn't pass the CI.
On second attempt, I used lerna to run `npm audit fix` across all the project packages.


mynameistechno:
Came here for this, thanks for fixing.

coreyfarrell:
You did not read my comment on the last PR?

> The dependency "handlebars": "^4.1.0" will allow the non-vulnerable version to be installed so the issue is not here. The issue is that nyc@13.3.0 bundles dependencies. We're in the processes of releasing nyc 14.0.0 which will not include bundled dependencies and does not contain this vulnerability.
>
> Worth noting this is a false positive as istanbul-reports does not use untrusted templates.

nyc@14.0.0 is now released, please upgrade and regenerate your package-lock.json, the `npm audit` will be cleared.

adamkingit:
Sorry, I realized my original PR didn't pass the travis build, so went to fix it before I saw your PR comment. Thanks for reviewing so quickly.

PR #375, opened 2019-04-21
fix: if LEAST_UPPER_BOUND returns null, try GREATEST_LOWER_BOUND

bcoe:
I'm running into issues adding source-map support to c8, in which `LEAST_UPPER_BOUND` returns `mapping.source === null`.

If I loosen up our check so that it falls back to the default behavior of `GREATEST_LOWER_BOUND` in this edge-case, c8's remapping works like a charm, see:

Screen Shot 2019-04-20 at 11 46 44 PM

----

@loganfsmyth can you clarify for me why this change might address my issues? I understand that they underlying cause is most likely that V8's coverage ranges don't directly map to the mappings generated by TypeScript ... but why would `LEAST_UPPER_BOUND` be returning `null`? 

Reading the documentation, shouldn't it simply return the next closest element that's a greater line number than the element I'm remapping -- I guess I don't fully understand the scenarios that result in `LEAST_UPPER_BOUND` or `GREATEST_LOWER_BOUND` returning `null`.

### DIFF
--- a/packages/istanbul-lib-source-maps/lib/get-mapping.js
+++ b/packages/istanbul-lib-source-maps/lib/get-mapping.js
@@ -74,22 +74,20 @@ function originalEndPositionFor(sourceMap, generatedEnd) {
 
 /**
 * Attempts to determine the original source position, first
-* trying LEAST_UPPER_BOUND which returns the closest
-* element larger than the element we were searching for. If this fails, we next
-* try GREATEST_LOWER_BOUND, which returns the closest element smaller than
-* the element being searched for.
+* returning the closest element to the left (GREATEST_LOWER_BOUND),
+* and next returning the closest element to the right (LEAST_UPPER_BOUND).
 */
 function originalPositionTryBoth (sourceMap, line, column) {
   let mapping = sourceMap.originalPositionFor({
       line: line,
       column: column,
-      bias: LEAST_UPPER_BOUND
+      bias: GREATEST_LOWER_BOUND
   });
   if (mapping.source === null) {
     mapping = sourceMap.originalPositionFor({
         line: line,
         column: column,
-        bias: GREATEST_LOWER_BOUND
+        bias: LEAST_UPPER_BOUND
     });
   }
   return mapping;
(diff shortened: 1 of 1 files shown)


packages/istanbul-lib-source-maps/lib/get-mapping.js
```diff
@@ -98,7 +121,11 @@ function getMapping(sourceMap, generatedLocation, origFile) {
         return null;
     }
 
-    const start = sourceMap.originalPositionFor(generatedLocation.start);
+    const start = originalPositionTryBoth(
+      sourceMap,
+      generatedLocation.start.line,
+      generatedLocation.start.column
+    );
     let end = originalEndPositionFor(sourceMap, generatedLocation.end);
```
bcoe: I believe, in the case of my testing with `c8`, it was the `GREATEST_LOWER_BOUND` in the `originalEndPositionFor` that was resulting in `null` data; opting for `LEAST_UPPER_BOUND` works like a charm ... but I don't fully understand why both wouldn't return an element.

bcoe:
I dug into the `source-map` codebase, and can speak better to why this logic is necessary:

If you have `GREATEST_LOWER_BOUND` configured, `source-map` returns `-1` rather than the closest element, if there's no element smaller than the node we're searching for in the tree:

```js
    // we are in termination case (3) or (2) and return the appropriate thing.
    if (aBias == exports.LEAST_UPPER_BOUND) {
      return mid;
    } else {
      return aLow < 0 ? -1 : aLow;
    }
```

inversely, if `LEAST_UPPER_BOUND` is configured, `source-map` returns `-1` if there is no element greater than the element being searched for:

```js
    if (aBias == exports.LEAST_UPPER_BOUND) {
      return aHigh < aHaystack.length ? aHigh : -1;
    } else {
      return mid;
    }
```

---

By trying a search for `LEAST_UPPER_BOUND` and `GREATEST_LOWER_BOUND` what we're essentially doing is ensuring that we get the `mid` point returned to us. This behavior seems more reasonable than completely falling over when generated ranges and source map ranges don't line up well -- as is the case with `c8`.

There's a similar conversation happening here about trying both approaches:

https://github.com/mozilla/source-map/pull/376

PR #376, opened 2019-04-21
Run all files through prettier

lukeapage:
prettier is configured but not run on every file.

So.. I'd suggest either adding filetypes to prettier ignore so they aren't formatted.. or format them.

I've gone with the latter on this PR.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,42 +1,42 @@
 {
-  "extends": [
-    "eslint:recommended",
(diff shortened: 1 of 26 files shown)


lukeapage:
> packages/istanbul-lib-instrument/api.md is generated by the `documentation` package so I think it needs to be ignored.

done

coreyfarrell:
@lukeapage thanks, merged.

PR #377, opened 2019-04-21
Make the html report into a class and allow the breadcrumb to be overridden

lukeapage:
This is the only change required to allow https://github.com/istanbuljs/istanbuljs/pull/345 - because that uses the existing report to generate the file nodes, which include a breadcrumb.. so it overrides the report breadcrumb function.

I thought I would modernize it to a class at the same time, since thats whats been suggested of the classes in #345


coreyfarrell:
@lukeapage thanks for helping us modernize the code!

PR #378, opened 2019-04-22
fix: Deprecate istanbul-api.

### LINKED ISSUE #321: Explore deprecation of istanbul-api
I'd like for us to consider deprecating `istanbul-api` in favor of consumers directly using the implementation modules (`istanbul-lib-report`, etc).  The motivation is `source-map@0.7.3` - the async constructor requires breaking API changes to `istanbul-lib-source-maps` and I'm not sure how to make `istanbul-api` work with it.  One idea I have is to publish `istanbul-api@3.0.0` and immediate deprecate that version only.  This way we discourage new modules from using istanbul-api but we don't cause a deprecation warning for the existing major users.  When we drop node.js 6 and upgrade `source-map` we would not update `istanbul-api` to be compatible.

According to npmjs.com `istanbul-api` was downloaded 5.1 million times last week.  `jest-cli` was downloaded 4.5 million times.  `jest-cli` is being split up so the `istanbul-api` dependency is moving to `@jest/reporters`.  `createReporter` is the only thing jest uses from `istanbul-api` so I'm hoping we can convince them to use `istanbul-lib-report` directly.

Another large user is `karma-coverage-istanbul-reporter` at 700k downloads last week.  `fusion-cli` had 15k last week, I'm not seeing any other dependents with a large number of downloads.

About the missing 200k `istanbul-api` downloads, I'm guessing `jest-cli@14` and below is still getting significant downloads, that version uses the original `istanbul` module.

coreyfarrell:
BREAKING CHANGE: The istanbul-api module is now deprecated.  Any modules
which use istanbul-api should migrate.

Fixes #321

### DIFF
--- a/packages/istanbul-api/README.md
+++ b/packages/istanbul-api/README.md
@@ -3,4 +3,12 @@
 [![Greenkeeper badge](https://badges.greenkeeper.io/istanbuljs/istanbul-api.svg)](https://greenkeeper.io/)
 [![Build Status](https://travis-ci.org/istanbuljs/istanbul-api.svg?branch=master)](https://travis-ci.org/istanbuljs/istanbul-api)
 
-High-level API for istanbul.
+Deprecated high-level API for istanbul.
+
+Remaining users of this module should migrate:
+
+-   istanbul-lib-instrument or babel-plugin-istanbul - modify sources to include coverage counters
+-   istanbul-lib-coverage - merge coverage results from multiple tests into a single dataset for reporting
+-   istanbul-lib-source-maps - process source-maps before generating reports
+-   istanbul-lib-report - the reporting system
+-   istanbul-reports - the actual reports (text-summary report, html report, etc)


coreyfarrell:
The plan is that after merging this we will perform a semver-major release of istanbul-api (3.0.0) to npm, then immediately deprecate that version.  Once that is completed a follow-up will delete packages/istanbul-api and remove the reference to it from `greenkeeper.json` and `.eslintignore`.

Users of packages which depend on istanbul-api@2 will not get any warning, maintainers will get the deprecation notice if they try to upgrade.

lukeapage:
IMO the reporter part of api should be moved to istanbul-lib-report to provide a easy way to do reporting that gives us the flexibility to change. It should probably just be the interface to that module and then not reference the reports - as it is, to use these 2 modules you have to know quite a few very specific implementation details and it ties istanbul up to breaking changes if any implementation details change or new reports come in that aren't strict visitors (like html-spa)

coreyfarrell:
@lukeapage I appreciate your concerns and I'm absolutely open to suggestions to improve the istanbul-lib-report interface (sorry for the slow response past couple weeks have been difficult for me).  As mentioned on the Modern HTML PR I'm open to semver-major actions to istanbul-lib-report / istanbul-reports which would shift some of the API burden away from the code running reports.

Just so you are aware part of the problem is a lack of knowledge about istanbul-api among maintainers.  It would help me to know of specific istanbul-api features which you would like to see supported by istanbul-lib-report.  Sorry if this sounds like a silly question but I've never used istanbul-api so I can't understand the value of it.

lukeapage:
I’ve only ever used the api lib to make reporters straight forward so my concerns are satisfied. I could have a go at a new api proposal if I get some time.

coreyfarrell:
@lukeapage please hold off for now, chances are high that I will have a PR posted this weekend to improve istanbul-lib-report (I already have some WIP).  I'll tag you on that PR when it's created.

lukeapage:
Sounds good!

PR #379, opened 2019-04-23
feat: Enable classProperties and classPrivateProperties parsers.

coreyfarrell:


coreyfarrell:
@bcoe I'd like to make sure this is in the `next` release now that node.js 12 is out and testing can be enabled.

packages/istanbul-lib-instrument/src/instrumenter.js
```diff
@@ -26,11 +26,8 @@ export function defaultOpts() {
         plugins: [
             'asyncGenerators',
             'bigInt',
-            /* Verify and add upon release of node.js 12:
-             * classProperties
-             * classPrivateProperties
-             * classPrivateMethods
```
coreyfarrell: It seems classPrivateMethods are not supported in node.js v12.0.0.  Will be keeping an eye on the v8 release blog and add plugins when appropriate.

packages/istanbul-lib-instrument/test/util/guards.js
```diff
@@ -18,6 +18,14 @@ export function isYieldAvailable() {
     return tryThis('function *foo() { yield 1; }', 'yield');
 }
 
+export function isClassPropAvailable() {
+    return tryThis('class Foo { a = 1; }', 'class property');
+}
+
+export function isClassPrivatePropAvailable() {
+    return tryThis('class Foo { #a = 1; }', 'class private property');
```
bcoe: oh my goodness, I'm unreasonably excited about this feature.

coreyfarrell:
> :+1: this is awesome! private class properties are available now?

Yes in Chrome 74 and Node.js 12.  Private getters/setters `get #foo() { return 'something'; }` and private methods `#foo() {}` are not yet available, hopefully will be implemented in v8-7.5 so we get it during node.js 12 lifecycle.

PR #381, opened 2019-04-28
fix: Also ignore *.cjs, *.mjs and .*ts by default

shinnn:
When `--experimental-modules` flag is enabled, those extensions are used to specify whether the script is CommonJS or ES module explicitly. https://nodejs.org/api/esm.html

### DIFF
--- a/packages/test-exclude/index.js
+++ b/packages/test-exclude/index.js
@@ -177,8 +177,8 @@ exportFunc.defaultExclude = [
     'coverage/**',
     'packages/*/test/**',
     'test/**',
-    'test{,-*}.{,c,m}js',
-    '**/*{.,-}test.{,c,m}js',
+    'test{,-*}.{js,cjs,mjs}',
+    '**/*{.,-}test.{js,cjs,mjs}',
     '**/__tests__/**',
     `**/{${devConfigs.join()}}.config.js`
 ];
--- a/packages/test-exclude/index.js
+++ b/packages/test-exclude/index.js
@@ -177,8 +177,8 @@ exportFunc.defaultExclude = [
     'coverage/**',
     'packages/*/test/**',
     'test/**',
-    'test{,-*}.{js,cjs,mjs}',
-    '**/*{.,-}test.{js,cjs,mjs}',
+    'test{,-*}.{js,cjs,mjs,ts}',
+    '**/*{.,-}test.{js,cjs,mjs,ts}',
     '**/__tests__/**',
     `**/{${devConfigs.join()}}.config.js`
 ];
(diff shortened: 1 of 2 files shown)


packages/test-exclude/test/test-exclude.js
```diff
@@ -211,10 +223,10 @@ describe('testExclude', () => {
             'coverage/**',
             'packages/*/test/**',
             'test/**',
-            'test{,-*}.js',
-            '**/*{.,-}test.js',
+            'test{,-*}.{,c,m}js',
+            '**/*{.,-}test.{,c,m}js',
             '**/__tests__/**',
-            '**/{ava,babel,jest,nyc,rollup,webpack}.config.js'
+            '**/{ava,babel,jest,nyc,rollup,webpack}.config.{,c,m}js'
```
shinnn: These libraries don't allow config files to have `cjs` or `mjs` extension yet, I made test-exclude support them for future proof.

packages/test-exclude/index.js
```diff
@@ -177,10 +177,10 @@ exportFunc.defaultExclude = [
     'coverage/**',
     'packages/*/test/**',
     'test/**',
-    'test{,-*}.js',
-    '**/*{.,-}test.js',
+    'test{,-*}.{,c,m}js',
```
coreyfarrell: For this (and the next line) I'd prefer that we write out each extension: `{js,cjs,mjs}`.  test-exclude is not a web module so the 4 character savings is not that big and I think this makes it much easier to read.

@bcoe Should we be adding `ts` to the list of excluded tests?

packages/test-exclude/test/test-exclude.js
coreyfarrell: I don't think we should add these options until/unless the projects actually support them, or at least indicate plans to support them.  As an example `ava.config.js` is loaded using the `esm` module, the `.js` extension is explicit in documentation and code.

shinnn: https://github.com/istanbuljs/istanbuljs/pull/381/commits/ed82e4a2af95b9e819437dca5e70ab39a7b6eb0b

packages/test-exclude/index.js
shinnn: >  I'd prefer that we write out each extension: `{js,cjs,mjs}`. 

5cf2388b74cc6d12232b4fd497bb49b04e07487e

coreyfarrell: I spoke with @bcoe in chat, would you mind adding `ts` to the test file extensions?

shinnn: 8cba44077f9a5135d50ef28b5d70f41b964f1067

Issue #66, opened 2017-06-12
Packages is N/A with cobertura report if test files is under one folder 


flyHe:
moved from https://github.com/istanbuljs/nyc/issues/598#issuecomment-306957049

Please use the template provided below, when reporting bugs:
-----

## Expected Behavior

Packages should be 100% with cobertura report if test files is under one folder
The right cobertura xml file should be as below (generated by istanbul directly):

<img width="1087" alt="screen shot 2017-06-07 at 4 39 47 pm" src="https://user-images.githubusercontent.com/10543919/26869803-4fa56cb0-4ba0-11e7-9a78-b27ffa459ffe.png">

## Observed Behavior

Packages is N/A with cobertura report if test files is under one folder

<img width="625" alt="screen shot 2017-06-07 at 4 21 36 pm" src="https://user-images.githubusercontent.com/10543919/26869880-a632f91c-4ba0-11e7-912e-7ed2b67dea2a.png">

<img width="1093" alt="screen shot 2017-06-07 at 4 39 54 pm" src="https://user-images.githubusercontent.com/10543919/26869811-556a3888-4ba0-11e7-8080-2bcba6889eb0.png">

## Forensic Information

**Operating System:** OS X Sierra 10.12.5
**Environment Information:** node 6.10.3, npm 3.10.10

amerryma:
Seeing this as well! 👍

bcoe:
@amerryma @flyHe I'm not too familiar with the cobertura test reporter. Would either of you be able to provide an example repo that reproduces this issue? ideally with expected output, vs. observed output?

At which point, happy to start digging into the issue (or to accept any patches you make to address the issue, if you feel like doing some digging yourselves).

flyHe:
@bcoe I have created an example repo in https://github.com/flyHe/testSample

ivailop:
Just FYI, I believe this issue is exactly the same as https://github.com/istanbuljs/istanbuljs/issues/14. Notice there was also a similar Clover report problem - https://github.com/istanbuljs/istanbuljs/issues/13.

I'm sorry I had no time confirming either of them since they were considered to be fixed.

bcoe:
@flyHe thank you I appreciate the sample repo; if anyone beats me to it, would love help digging into the underlying issue with the cobertura reporter.

jhillacre:
I can get a valid cobertura,xml report generated by commenting out these lines:
[...]

PR #382, opened 2019-04-30
fix(istanbul-reports): Remove isRoot check causing incorrect report formatting (#66).

coderica:
This issue has been hanging out for a long while and seems to have effected a lot of other people. I thought I had discovered a fix by just ensuring my source files were not all in one directory but then the  problem popped up again.

I could not discover a good reason why this check exists, more specifically why it exists only for these tags. In fact, all it seems to do is break the report format when its generated, which means its no good for the Jenkins API :(

This bug has blocked my work several times now so I figured I'd try contributing the fix.

### DIFF
--- a/packages/istanbul-reports/lib/cobertura/index.js
+++ b/packages/istanbul-reports/lib/cobertura/index.js
@@ -57,9 +57,6 @@ CoberturaReport.prototype.writeRootStats = function(node) {
 };
 
 CoberturaReport.prototype.onSummary = function(node) {
-    if (node.isRoot()) {
-        return;
-    }
     const metrics = node.getCoverageSummary(true);
     if (!metrics) {
         return;
@@ -72,10 +69,7 @@ CoberturaReport.prototype.onSummary = function(node) {
     this.xml.openTag('classes');
 };
 
-CoberturaReport.prototype.onSummaryEnd = function(node) {
-    if (node.isRoot()) {
-        return;
-    }
+CoberturaReport.prototype.onSummaryEnd = function() {
     this.xml.closeTag('classes');
     this.xml.closeTag('package');
 };


coderica:
fixed my typo in the commit message

coderica:
I have tested it locally on my project. The isRoot check really does not seem to accomplish anything but break the expected report format. Removing it enables the report to be generated in its entirety.

coreyfarrell:
Thanks for the contribution and for confirming that you tested!

coderica:
when will this change be released?

coderica:
@coreyfarrell when can i expect this change to be released? I'll need to followup with an upgrade in Jest

coderica:
also will https://github.com/istanbuljs/istanbuljs/issues/66 be closed automatically?

coreyfarrell:
@coderica please be patient, it'll should be some time this week.  I'm pretty sure jest will not need to make any changes as the new version of istanbul-reports will be allowed by the semver-range jest uses.  Once this is released you should be able to just run `npm update` to update your package-lock.json if you have one.

bcoe:
@coderica thanks for this fix 🎉 it's awesome to squash an old bug like this.

coreyfarrell:
This is now released via `istanbul-reports@2.2.5`.

bcoe:
@coderica, @coreyfarrell and I know basically nothing about Cobertura, and the problem with this PR wasn't human error, it was lack of test cases.

Would you be interested in working with us to get this over the finish line:

1. we can rewrite the test case shared [here](https://github.com/istanbuljs/istanbuljs/issues/384#issuecomment-489336932) to work in our test suite.
2. we can try to add a test case for the behavior you're fixing.
3. we can release your work to the `next` tag and have folks test it before it goes to `latest` ... if we find more edge-cases, we can just add more tests.

Appreciate your contributions, sorry that OSS can be frustrating sometimes 😝

matthewjohnson42:
Thank you for this fix @coderica

PR #383, opened 2019-05-01
fix typescript config docs

JaKXz:


JaKXz:
@coreyfarrell I'll let you merge and release this - I'd say it's a patch fix to the nyc-config-typescript package so that the docs on npm.im/@istanbuljs/nyc-config-typescript are updated

coreyfarrell:
@JaKXz any change to one of the monorepo packages ends up being a `patch` release by default, unless the commit is tagged `feat:` or the content has `BREAKING CHANGE:` (then it's minor or major respectively).  I'll probably tag this commit with `docs:` because it doesn't need to be listed in the changelog.

PR #388, opened 2019-05-04
chore: Use class syntax in istanbul-lib-report

coreyfarrell:


packages/istanbul-lib-report/test/tree.test.js
```diff
@@ -9,31 +8,32 @@ const BaseNode = t.Node;
 const Visitor = t.Visitor;
 const CompositeVisitor = t.CompositeVisitor;
 
-function TestNode(name) {
-    BaseNode.call(this);
-    this.name = name;
-    this.parent = null;
-    this.children = [];
-}
+class TestNode extends BaseNode {
```
coreyfarrell: This had to be converted because it's no longer valid to run `BaseNode.call(this)` as BaseNode is a class (so it must be called with `new`).  This is not a breaking change as `BaseNode` is not exposed outside this module (`lib/tree.js` is not part of the public API).

packages/istanbul-lib-report/lib/tree.js
```diff
@@ -23,8 +22,10 @@ const util = require('util');
...
  */
-function Visitor(delegate) {
-    this.delegate = delegate;
+class Visitor {
+    constructor(delegate) {
+        this.delegate = delegate;
+    }
 }
 
 ['Start', 'End', 'Summary', 'SummaryEnd', 'Detail'].forEach(k => {
```
JaKXz: yikes :\

Can we do this with `Object.defineProperty` instead at least?

packages/istanbul-lib-report/lib/tree.js
```diff
@@ -59,97 +62,80 @@ util.inherits(CompositeVisitor, Visitor);
...
-Node.prototype.getParent = function() {
-    throw new Error('getParent must be overridden');
-};
-
-/* istanbul ignore next: abstract method */
-Node.prototype.getChildren = function() {
-    throw new Error('getChildren must be overridden');
-};
+class Node {
+    /* istanbul ignore next: abstract method */
```
JaKXz: is this really an abstract method?

packages/istanbul-lib-report/lib/context.js
```diff
@@ -41,12 +42,79 @@ function mergeWatermarks(specified, defaults) {
...
-    this.watermarks = mergeWatermarks(opts.watermarks, watermarks.getDefault());
-    this.sourceFinder = opts.sourceFinder || defaultSourceLookup;
-    this.data = {};
+class Context {
+    constructor(opts) {
+        opts = opts || {};
+        this.dir = opts.dir || 'coverage';
+        this.watermarks = mergeWatermarks(
+            opts.watermarks,
+            watermarks.getDefault()
```
JaKXz: can `mergeWatermarks` make this an optional argument? 

on second thought that might not be clearer...

packages/istanbul-lib-report/lib/summarizer.js
```diff
@@ -4,98 +4,99 @@
...
+        child.parent = this;
+        this.children.push(child);
+    }
 
-ReportNode.prototype.asRelative = function(p) {
-    /* istanbul ignore if */
-    if (p.substring(0, 1) === '/') {
-        return p.substring(1);
+    asRelative(p) {
+        /* istanbul ignore if */
```
JaKXz: why are we ignoring this if?

packages/istanbul-lib-report/lib/file-writer.js
```diff
@@ -2,193 +2,196 @@
...
+        fs.closeSync(this.fd);
+    }
+}
 
-FileContentWriter.prototype.close = function() {
-    fs.closeSync(this.fd);
-};
+// allow stdout to be captured for tests.
+let capture = false;
+let output = '';
```
JaKXz: I don't really understand this change, would you mind explaining it? This looks like "state" to me and shouldn't that be in the [base] class?

packages/istanbul-lib-report/lib/tree.js
coreyfarrell: Currently it is.  A follow-up patch will actually make a constructor for `class Node` that will accept a `root` argument so `getRoot` will no longer be abstract.

packages/istanbul-lib-report/lib/file-writer.js
coreyfarrell: This is not a "change", just moved code.

Really IMO this `capture` functionality should be removed, the test should just replace `process.stdout`.  I'd rather avoid adding complexity to this PR with a change like this.

packages/istanbul-lib-report/lib/context.js
coreyfarrell: Possibly worth doing in a follow-up.

packages/istanbul-lib-report/lib/summarizer.js
coreyfarrell: I just checked, the `if` path is not hit.  I'll remove the ignore though as I don't think it's impossible to test for this.

packages/istanbul-lib-report/lib/tree.js
coreyfarrell: Done

packages/istanbul-lib-report/lib/tree.js
JaKXz: Interesting... it has a defined body though, so isn't it not-abstract by definition?

coreyfarrell: If you mean abstract like C++ classes have abstract methods, that's not a feature of JS.  In any case these "abstract" functions are replaced in all cases so it could probably be deleted since these are internal classes that are overridden and the "abstract" functions are implemented in child classes.

JaKXz: Yeah sorry it's not "strictly" abstract - anyway, I agree, I think it should just be deleted... but maybe as a follow up.

JaKXz:
@coreyfarrell feel free to merge when you're ready 👍

PR #389, opened 2019-05-04
chore: Internal refactoring of istanbul-lib-report

coreyfarrell:
* Drop "abstract" functions
* Rename Node and Tree to BaseNode and BaseTree
* Create ReportNode.createRoot static method
* Replace treeFor with ReportTree class
* Simplify multiple internal `summarizer.js` functions
* Implement getRoot in BaseTree

Ref #390

---

Will be additional follow-up but this is likely the last of the non-breaking changes for istanbul-lib-report.

CC @lukeapage

### DIFF
--- a/packages/istanbul-lib-report/lib/file-writer.js
+++ b/packages/istanbul-lib-report/lib/file-writer.js
@@ -8,20 +8,11 @@ const mkdirp = require('make-dir');
 const supportsColor = require('supports-color');
 
 /**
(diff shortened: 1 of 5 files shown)


lukeapage:
I reviewed it and didn't have any comments.. there's lots of improvement to the code structure and better code reuse. I only spent 30 mins on it though, I didn't work through everything.

PR #396, opened 2019-05-09
chore: Delete istanbul-api

coreyfarrell:
Do not merge this until after we've done a release from lerna - we need istanbul-api@3.0.0 to be published to npm so it can be deprecated.  The main point of posting this now is to see how travis reacts.


coreyfarrell:
This appears to have very negative effect on our reported coverage of istanbul-reports, small decreases in other packages.  I intentionally say "reported coverage" because I think these numbers were false.  For example master currently shows 100% coverage for all metrics on the `lcov` report, 0% coverage after this PR.  Problem is the istanbul-api tests never actually tested the contents of `lcov` reports.  It only verified that the report ran without throwing, created the expected file and that the file was non-empty.

So I think `master` is showing false coverage and we need to seriously expand our istanbul-reports testing.

coreyfarrell:
Additional commit ensures that coverage for `istanbul-lib-instrument` only includes tests written specifically for `istanbul-lib-instrument`.  This will help avoid false coverage.

coreyfarrell:
istanbul-api is getting in the way of the work I'm doing on reporting so I'm merging this now.  I already manually published istanbul-api@3.0.0 to `next`.  Using lerna really wasn't an option because we need istanbul-api@3.0.0 to depend on the existing versions of other monorepo packages.  The istanbul-api@3.0.0 is not tagged because that would interfere with the next `lerna publish`.  Will try remembering to add that tag after the next full monorepo publish.

PR #397, opened 2019-05-10
chore: Isolate coverage reporting per package

coreyfarrell:
Test coverage should only count if we're directly testing the result of
something.  This runs nyc per package then produces combined reporting.

---

This effectively drops coverage from istanbul-api because I want to avoid any further commits to that package (other than deleting it).  This commit was part of #396 but I want to push it ahead.

This causes the reported coverage to drop significantly but this is more valid.

### DIFF
--- a/packages/nyc-config-hook-run-in-this-context/package.json
+++ b/packages/nyc-config-hook-run-in-this-context/package.json
@@ -7,7 +7,7 @@
     "index.js"
   ],
   "scripts": {
-    "test": "mocha"
+    "test": "nyc --nycrc-path=../../monorepo-per-package-nycrc.json mocha"
   },
   "repository": {
     "type": "git",
@@ -31,6 +31,11 @@
   },
   "dependencies": {
     "semver": "^6.0.0"
+  },
+  "devDependencies": {
+    "chai": "^4.2.0",
+    "mocha": "^6.1.4",
+    "nyc": "^14.1.0"
   },
   "engines": {
     "node": ">=6"
--- a/monorepo-merge-reports.js
+++ b/monorepo-merge-reports.js
@@ -1,0 +1,33 @@
+#!/usr/bin/env node
+'use strict';
+
+const path = require('path');
+const { spawnSync } = require('child_process');
+
+const rimraf = require('rimraf');
+const makeDir = require('make-dir');
+const glob = require('glob');
+
+process.chdir(__dirname);
+rimraf.sync('.nyc_output');
+makeDir.sync('.nyc_output');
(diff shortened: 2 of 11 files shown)


nycrc-package.json
```diff
@@ -0,0 +1,6 @@
+{
```
JaKXz: can this file be renamed to be a bit clearer? or just published as `@istanbuljs/monorepo-per-package-coverage` or something and then used in the `extends` keys for each of the packages?

packages/nyc-config-hook-run-in-this-context/package.json
```diff
@@ -32,6 +32,11 @@
   "dependencies": {
     "semver": "^6.0.0"
   },
+  "devDependencies": {
+    "chai": "^4.2.0",
+    "mocha": "^6.1.4",
+    "nyc": "^14.1.0"
```
JaKXz: just a quick question /curiousity, remind me how lerna's hoisting works again? it just worries me having so many copies of these lines with nyc, mocha, and chai that could drift...

nycrc-package.json
coreyfarrell: I'd rather not publish to npm since this is not useful outside our monorepo.  Maybe rename the file to monorepo-per-package-nycrc.json would be better?

packages/nyc-config-hook-run-in-this-context/package.json
coreyfarrell: All of these packages are deduplicated to the root node_modules but a symlink is created for the binaries.  So `packages/nyc-config-hook-run-in-this-context/node_modules/.bin/nyc` will be a symlink to the binary in the root node_modules.

Any difference in raw range will cause warning on `npm install`, so if one package depends on `^14.1.0` and another depends on `^14.1.1` it will warn.  These are all managed by greenkeeper anyways so for example all packages will always be upgraded to nyc ^15.0.0 at the same time.

nycrc-package.json
JaKXz: Why not though? It could be!

JaKXz: An advanced example of `nyc merge` would be a good place to start.

coreyfarrell: OK but do you mind if this is done in follow-up?  I don't want to hold up this change while we figure out the documentation, naming of the new package, etc.  I think merge-reports.js would belong in the new package as an exposed bin (nyc-merge-monorepo-reports maybe), probably make it more generic by accepting glob(s) in the argument.

coreyfarrell:
I gave the files better names and created an #399 for the follow-up.

PR #398, opened 2019-05-10
feat!: Remove configuration loading functionality

### LINKED ISSUE #392: Remove config loading from test-exclude?
The purpose of this proposal is to simplify `test-exclude` and remove `read-pkg-up` and `require-main-filename` dependencies.  nyc doesn't (can't) use this functionality.

This was added a very long time ago in istanbuljs-archived-repos/test-exclude#1.  @bcoe do you remember if this was added for a specific request?  CC @JaKXz

coreyfarrell:
Fixes #392

### DIFF
--- a/packages/test-exclude/README.md
+++ b/packages/test-exclude/README.md
@@ -16,28 +16,6 @@ if (exclude().shouldInstrument('./foo.js')) {
 }
 ```
 
-_you can load configuration from a key in package.json:_
-
-_package.json_
-
-```json
-{
-    "name": "awesome-module",
-    "test": {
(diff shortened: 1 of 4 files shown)


coreyfarrell:
This is something I'm not clear on conventional-commits, what is the appropriate tag when you're removing a feature?

coreyfarrell:
> I'm not sure either but it should be marked as BREAKING CHANGE of course

`feat!:` (the exclamation mark) does that with the current version of standard-version.  I'll double-check that it works under lerna before I merge this.

PR #415, opened 2019-06-05
Adds ts-node/register/transpile-only as require option

godu:
https://github.com/istanbuljs/nyc/issues/1124#issuecomment-499071658

When we enable `nyc.all` option, `nyc` need to transpile source in the same way as the test runner (mocha or ava in my case).

If we add `--require ts-node/register/transpile-only` as default configuration for typescript, `--all` will work out of the box.

### DIFF
--- a/packages/nyc-config-typescript/README.md
+++ b/packages/nyc-config-typescript/README.md
@@ -18,6 +18,7 @@ And write a `.nycrc` that looks like this:
 ```js
 {
     "extends": "@istanbuljs/nyc-config-typescript",
+
     // OPTIONAL if you want coverage reported on every file, including those that aren't tested:
     "all": true
 }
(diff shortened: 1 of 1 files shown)


coreyfarrell:
I don't think this can be in the preset as some people might want to use `ts-node/register` (not transpile-only).

Maybe it would be better to just update the documentation to suggest the `require` in nyc config instead of mocha config? @JaKXz this more your area so I'll defer to your judgement for how to handle this.

godu:
Do someone realy want to use `ts-node/register` instead of `ts-node/register/transpile-only` ?
We only only the transpiler part of typescript here. It will be more efficient I think.

What uses `--require` apart from to transpile source when `--all` is enabled ?

coreyfarrell:
> What uses --require apart from to transpile source when --all is enabled ?

This also causes sources to be processed when actually running the tests.

JaKXz:
@godu the reason I'm against this is that the default `mocha` setup [does not need nyc to require `ts-node` at all](https://istanbul.js.org/docs/tutorials/typescript/) - unless this has changed recently, I would rather leave it so that people don't have to do

```js
"extends": "@istanbuljs/nyc-config-typescript",
"require": []
```

godu:
@JaKXz seems legit

I rework my PR to add some documentation about require. Tell me if you have any others requirements.

JaKXz:
@godu I appreciate that, but I don't think this adds any value to the package README, since it already mentions an `.nycrc` file.

PR #419, opened 2019-06-08
feat: ignore files under test"s" directories by default

shinnn:
I often use the name `tests` rather than `test` when the directory contains multiple test files.

### DIFF
--- a/packages/test-exclude/default-exclude.js
+++ b/packages/test-exclude/default-exclude.js
@@ -4,8 +4,8 @@ const devConfigs = ['ava', 'babel', 'jest', 'nyc', 'rollup', 'webpack'];
 
 module.exports = [
     'coverage/**',
-    'packages/*/test/**',
-    'test/**',
+    'packages/*/test{,s}/**',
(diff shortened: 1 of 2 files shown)


coreyfarrell:
Does any widely used testing framework look for tests in `tests/**` by default?

shinnn:
At least [AVA](https://github.com/avajs/ava), which has received over 16,000 stars on GitHub, is downloaded over 140,000 times every week and [is already taken into consideration by test-exclude](https://github.com/istanbuljs/istanbuljs/blob/c4e8b8e640acba43ecbe26ca3032a951de3ac3a6/packages/test-exclude/default-exclude.js#L3), picks `tests/**` up by default.

```console
$ cat tests/example.js
import test from 'ava';
test('Example', t => t.pass('Hi'));

$ ava

  1 test passed
```

bcoe:
@coreyfarrell I'm open to this addition if you don't have any objections.

coreyfarrell:
This works for me, thanks for contributing.

PR #425, opened 2019-06-21
Add example usage to istanbul-lib-report

lukeapage:


packages/istanbul-lib-report/README.md
```diff
@@ -3,4 +3,23 @@
 [![Greenkeeper badge](https://badges.greenkeeper.io/istanbuljs/istanbul-lib-report.svg)](https://greenkeeper.io/)
 [![Build Status](https://travis-ci.org/istanbuljs/istanbul-lib-report.svg?branch=master)](https://travis-ci.org/istanbuljs/istanbul-lib-report)
 
-Core reporting utilities for istanbul. Docs to follow.
+Core reporting utilities for istanbul.
+
+## example usage
```
coreyfarrell: Prefer `## Example Usage` title case for this

packages/istanbul-lib-report/README.md
```diff
@@ -3,4 +3,23 @@
...
+
+```
```
coreyfarrell: Please use the `js` tag for this code block so we get syntax highlighting.

packages/istanbul-lib-report/README.md
```diff
@@ -3,4 +3,23 @@
...
+const libReport = require('istanbul-lib-report');
+
+const context = libReport.createContext({
+  dir: 'source/dir',
```
coreyfarrell: This isn't the source dir, it's the report output directory.

coreyfarrell:
> Maybe another source for good examples would be jest's codebase, for inspiration

jest is currently not a good place to look.  jest is using the current version of istanbul modules.  index.js on istanbuljs/nyc#1134 is probably the only code that is using the `next` version of istanbul-lib-report.

lukeapage:
Fixed comments. I'm not saying this is perfect - its just I tried the alpha and updated my codebase and I ended up using the example on the refactor pr description, so I did this as I went along and made a PR so that _something_ exists ;)

packages/istanbul-lib-report/README.md
```diff
@@ -3,4 +3,28 @@
...
+## Example usage
+
+```js
+const libReport = require('istanbul-lib-report');
+const reports = require('istanbul-reports');
+
+// create a context for report generation
+const context = libReport.createContext({
+  dir: 'report/output/dir',
+  defaultSummarizer: 'summarizer name e.g. nested/flat/pkg',
```
JaKXz: I think this should be a valid value, and have a comment describing what it is, so that people have an idea what they can copy/paste and then extend as necessary. I've been working a lot with Gatsby recently and their docs do a good job of this.

packages/istanbul-lib-report/README.md
```diff
@@ -3,4 +3,28 @@
...
 [![Build Status](https://travis-ci.org/istanbuljs/istanbul-lib-report.svg?branch=master)](https://travis-ci.org/istanbuljs/istanbul-lib-report)
 
-Core reporting utilities for istanbul. Docs to follow.
+Core reporting utilities for istanbul.
+
+## Example usage
+
+```js
+const libReport = require('istanbul-lib-report');
+const reports = require('istanbul-reports');
```
JaKXz: Could we destructure these imports? That makes reading examples like this [from other docs] clearer for me.

packages/istanbul-lib-report/README.md
```diff
@@ -3,4 +3,28 @@
...
+// create a context for report generation
+const context = libReport.createContext({
+  dir: 'report/output/dir',
+  defaultSummarizer: 'summarizer name e.g. nested/flat/pkg',
+  watermarks: configWatermarks,
+  coverageMap,
+})
+
+// create an instance of the relevant report class
+const report = reports.create('report/name e.g. json/html/html-spa/text', {
```
JaKXz: same as the `defaultSummarizer` comment.

packages/istanbul-lib-report/README.md
```diff
@@ -3,4 +3,28 @@
...
+```js
+const libReport = require('istanbul-lib-report');
+const reports = require('istanbul-reports');
+
+// create a context for report generation
+const context = libReport.createContext({
+  dir: 'report/output/dir',
+  defaultSummarizer: 'summarizer name e.g. nested/flat/pkg',
+  watermarks: configWatermarks,
+  coverageMap,
```
JaKXz: Variables like these last 2 I would like to see defined, so that devs know what needs to be passed here or how to get the coverageMap to this point.

packages/istanbul-lib-report/README.md
lukeapage: since the used variables are "create" and "createContext" and therefore to be clear the imports would need to be renamed, I feel like destructuring in this case makes the code less readable.

packages/istanbul-lib-report/README.md
lukeapage: I defined configWatermarks.
For coverageMap - its a ball of string.. I could include libCoverage and create a coverage map.. but that requires either passing an empty report or loading a report and merging it in.. where does it end? this is not at the top level - how to use all istanbul packages, this is how to use istanbul-reports. As the main readme states, most end users won't need this (though I am writing these docs because I partly disagree and think the main readme documentation is way to aggressive and unhelpful)

lukeapage: I defined it with a comment.

coreyfarrell:
@lukeapage thanks for contributing!

Hypnosphi:
Doesn't seem to work with `istanbul-lib-report@2.0.8` and  `istanbul-reports@2.2.6`:

```
TypeError: reports.create(...).execute is not a function
```

coreyfarrell:
@Hypnosphi that is true, these docs are for the 3.x versions (currently pre-release).

PR #428, opened 2019-06-22
Fix none report to not cause an exception

lukeapage:
the base class takes the default summarizer, so without this change it tries to pass an options object into the report base as a summarizer name.

### DIFF
--- a/packages/istanbul-reports/lib/none/index.js
+++ b/packages/istanbul-reports/lib/none/index.js
@@ -5,6 +5,10 @@
  */
 const { ReportBase } = require('istanbul-lib-report');
 
-class NoneReport extends ReportBase {}
+class NoneReport extends ReportBase {
+    constructor() {
+        super();
+    }
+}
 
 module.exports = NoneReport;


coreyfarrell:
@lukeapage I'm thinking maybe `ReportBase` should take an options argument using `constructor(options = {})` and look for `options.summarizer`?  This way the `ReportBase` class would take the same arguments as the actual reports.

CC @bcoe @JaKXz

lukeapage:
I think it would be a better fix to do that.

bcoe:
@coreyfarrell your recommendation sounds like a reasonable design to me :+1:

coreyfarrell:
@lukeapage do you have time to update this so the `ReportBase` constructor takes the options argument?

lukeapage:
@coreyfarrell sorry for the delay - my available time depends on whats left over after work and family, hence the bursts of activity. Thanks for getting this sorted.

coreyfarrell:
Completely understand.. family should come first and work is a means to that ends.

PR #429, opened 2019-06-23
Allow merge to work consistently on windows

lukeapage:
On windows, the default `npm test` doesn't work for me.. investigating I see no documentation that spawn sync works with running a command installed in a sub node_modules.

I'm not 100% sue, I just know that master doesn't work for me.

This issue was informative: https://github.com/nodejs/node/issues/8077

### DIFF
--- a/monorepo-merge-reports.js
+++ b/monorepo-merge-reports.js
@@ -14,7 +14,7 @@ makeDir.sync('.nyc_output');
 // Merge coverage data from each package so we can generate a complete report
 glob.sync('packages/*/.nyc_output').forEach(nycOutput => {
     const { status, stderr } = spawnSync(
-        'nyc',
+        path.join('node_modules', '.bin', 'nyc'),
         [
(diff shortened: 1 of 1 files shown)


coreyfarrell:
@lukeapage Sorry to comment on a merged PR, I was AFK for almost 2 weeks.

In theory the `path.join` should not be needed as npm scripts are run with `node_modules/.bin` in the PATH, this is documented at https://docs.npmjs.com/cli/run-script.html.  So it's not spawn sync that works with commands installed in node_modules, it's the fact that `npm test` is an alias to `npm run test`.  Did you test this with `shell: true` but without using `path.join`?

coreyfarrell:
Additional comment, we have travis-ci run tests in Windows and did not get any failure before this PR.  I wonder why we don't get errors and if we can do something to our travis configuration to make it run tests in a mode which more closely resembles your local environment.

PR #432, opened 2019-06-24
Force html-spa to always be nested

lukeapage:
As previously discussed, this overrides the default summarizer so html-spa is always nested.


coreyfarrell:
Thanks @lukeapage, I've merged this with the `chore:` heading as the html-spa report is brand new so although this is technically a fix I don't want to clutter the changelog.

PR #433, opened 2019-06-24
Attempt a test setup for html-spa

lukeapage:
I didn't write tests for html-spa because it was awkward to get it working.

1. running mocha with babel on all tests fails, with some weird class constructor message
2. running mocha seperately (as in this pr) causes code coverage to be lost because the nyc local folder gets cleared. I worked around this with a env variable from nyc's source.

Overall this pr works, but its a bit hacky but I'm lost how else to do it, so I'm putting this up for comment.


bcoe:
:wave: I don't think you should have to pull in babel-register to get tests running (and I'd rather not since it changes behavior quite a bit).

I think you just need to run `npm i` in the top level folder, and then you should be able to run:

`npm test` 

In the top level folder.

lukeapage:
It’s the imports es6 format that breaks.
But actually I could rewrite it as commonjs?

bcoe:
@lukeapage how much of a rewrite would using commonjs be? If it's only a couple changes, I'd have a preference for this over adding any complication to the test scripts.

lukeapage:
It wasn't too bad - the main hassle was switching bundlers because rollup doesn't support a entry commonjs file. But I think this is better.
The one test is very simple, but I'll build upon tests as any changes are made. I wish I'd done this step wayy back tbh.

bcoe:
@lukeapage I don't have time to code review this week unfortunately, I will try to carve out some time this weekend; please feel free to message me if I fail to do so.

packages/istanbul-reports/package.json
```diff
@@ -19,20 +19,18 @@
     "@babel/core": "^7.4.5",
     "@babel/preset-env": "^7.4.5",
     "@babel/preset-react": "^7.0.0",
+    "babel-loader": "^8.0.6",
     "chai": "^4.2.0",
+    "cross-env": "^5.2.0",
```
coreyfarrell: I'm unclear why this is needed?

packages/istanbul-reports/lib/none/index.js
```diff
@@ -5,6 +5,10 @@
  */
 const { ReportBase } = require('istanbul-lib-report');
 
-class NoneReport extends ReportBase {}
+class NoneReport extends ReportBase {
```
coreyfarrell: Please undo this change as the ReportBase class has been fixed.

packages/istanbul-reports/package.json
lukeapage: Removed, not sure why I added (its needed if you need to set an env variable across environments)

packages/istanbul-reports/lib/none/index.js
lukeapage: done

lukeapage:
@coreyfarrell this is rebased and ready for merge afaik

bcoe:
@lukeapage :+1: excited to see this further along, will look soon myself too.

coreyfarrell:
Thanks @lukeapage, I'm planning to merge this weekend unless @bcoe or @JaKXz share any findings.

PR #438, opened 2019-06-25
html-spa: Use empty per metric and make the overall the worst of all metrics

lukeapage:
I've been discussing some things with @tisgro who is much better at ui/ux than me (he is working on a PR to make html-spa much prettier) and we came to some conclusions about a few elements of the report.

Firstly - when you have a file with multiple lines and no branches - it might be 0% covered, but because branches is 0/0 it is shown green - this is pretty misleading as you might be led to thinking nothing needs to be done on that file. The empty class is only used on each individual metric when all the metrics are empty. So this changes that, so each metric can be empty and if its empty, its shown white - basically ignore me.

Secondly - the colour applied to the filename comes from the statements - which isn't even an included metric! even if it was lines, its confusing because filters don't work only on lines - they work on the worst of all shown metrics. So I've changed the filename color to be based on the worst metric that isn't empty - this makes the colouring match the filtering, so you won't see green files when filtering for low (we still have a confusing problem with folders, but I'll tackle that seperately).

### DIFF
--- a/packages/istanbul-reports/test/fixtures/specs/maxcols.json
+++ b/packages/istanbul-reports/test/fixtures/specs/maxcols.json
@@ -20,15 +20,15 @@
                 "total": 0,
                 "covered": 0,
                 "skipped": 0,
-                "pct": 100,
-                "classForPercent": "high"
+                "pct": 0,
+                "classForPercent": "empty"
             },
             "functions": {
                 "total": 0,
                 "covered": 0,
                 "skipped": 0,
-                "pct": 100,
-                "classForPercent": "high"
+                "pct": 0,
+                "classForPercent": "empty"
             },
             "lines": {
                 "total": 18,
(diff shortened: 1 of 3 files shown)


packages/istanbul-reports/test/fixtures/specs/maxcols.json
```diff
@@ -80,206 +80,206 @@
         "/index.js": {
             "path": "/index.js",
             "statementMap": {
-              "0": {
```
lukeapage: sorry these changes here on came from prettier formatting the json.

lukeapage:
Preview of istanbul (note the empty look a bit more ugly, but I think should look better with redesign changes)

image

coreyfarrell:
@lukeapage when you have a chance could you rebase and resolve the conflicts?  Sorry this one dropped off the radar but based on your comments I wanted to get the redesign changes merged first.

lukeapage:
@coreyfarrell its rebased now

coreyfarrell:
@lukeapage thanks, merged.  Do we have remaining issues with the html-spa report which need to be resolved before the first official release?  The next release will be semver-major and I'm not ready yet, just wanted to get your opinion on the state of html-spa.

PR #439, opened 2019-06-25
html-spa: Update styles

tisgro:
This PR fixes several issues in the look-and-feel of the html-spa report:

🍱 **Reduce visual clutter:** The current design shows a lot of data with equal weight, making it quite difficult to get an overview of the coverage at a glance. The new design removes clutter, and prioritizes data visually, allowing the user to focus on the most important content.

💅 **Style toggle buttons:** The buttons to toggle view options are currently pretty rough, making it difficult to see the current state.

🥖 **Add file path breadcrumbs:** Add links to navigate to any part of the current path when drilling-down.

Preview: http://blonde-robot.surge.sh/


coreyfarrell:
This change will conflict with #433, I'm not sure which we want to push first?  Although I'd normally prefer to add testing first I suspect that fixing conflicts in #433 would be less work than fixing conflicts here?

CC @lukeapage

bcoe:
@tisgro @coreyfarrell @lukeapage I'm excited to have more folks contributing to the single page app design, would love to get this over the finish line ... @coreyfarrell it looks like there are only a few blockers on #433?

lukeapage:
I fixed up my comments.

Travis is on holiday a while, so if you want to merge this, I'll happily rebase #433 . I've now fixed the comments on that pr.

packages/istanbul-reports/lib/html-spa/assets/spa.css
```diff
@@ -74,183 +129,171 @@ a:hover {
...
+.coverage-summary td.pct {
+    font-weight: 400;
 }
-/* light yellow */
-.medium {
-    background: #fff4c2;
+
+.coverage-summary td.abs {
+    color: #666;
+    font-weight: 200;
```
coreyfarrell: This degrades the view, on my system it causes the render of these numbers to appear slightly pixelated.  I think leaving `font-weight` as is (normal) with the `font-size: 12px` adequately de-emphasizes the covered/total absolute counts while still allowing a smooth looking font render.  Tested by disabling this rule with the browser devtools.

tisgro: Fair point -- the fancy font weight was perhaps gilding the lily a bit ;)

I've removed the font weight on my branch now, so hopefully you can merge now.

PR #441, opened 2019-07-06
feat: ignore fixture files and directories by default

shinnn:
It’s enough common to exclude `fixture` and `fixtures` from coverage calculation targets.

### DIFF
--- a/packages/test-exclude/default-exclude.js
+++ b/packages/test-exclude/default-exclude.js
@@ -9,5 +9,6 @@ module.exports = [
     'test{,-*}.{js,cjs,mjs,ts}',
     '**/*{.,-}test.{js,cjs,mjs,ts}',
     '**/__tests__/**',
+    '**/fixture*{,/**}',
     `**/{${devConfigs.join()}}.config.js`
 ];
--- a/packages/test-exclude/test/test-exclude.js
+++ b/packages/test-exclude/test/test-exclude.js
@@ -53,6 +53,28 @@ describe('testExclude', () => {
             .should.equal(false);
     });
 
+    it('ignores fixture directories and files', () => {
+        exclude()
+            .shouldInstrument('./fixture/index.js')
+            .should.equal(false);
+
+        exclude()
+            .shouldInstrument('./fixtures/index.js')
+            .should.equal(false);
+
+        exclude()
+            .shouldInstrument('./fixture-some-suffix/index.js')
+            .should.equal(false);
+
+        exclude()
+            .shouldInstrument('./nested/fixtures/index.js')
+            .should.equal(false);
+
+        exclude()
+            .shouldInstrument('./fixtures-file.txt')
+            .should.equal(false);
+    });
+
     it('matches files in root with **/', () => {
         exclude()
             .shouldInstrument('__tests__/**')
@@ -247,6 +269,7 @@ describe('testExclude', () => {
             'test{,-*}.{js,cjs,mjs,ts}',
             '**/*{.,-}test.{js,cjs,mjs,ts}',
             '**/__tests__/**',
+            '**/fixture*{,/**}',
             '**/{ava,babel,jest,nyc,rollup,webpack}.config.js'
         ]);
     });
@@ -281,7 +304,7 @@ describe('testExclude', () => {
                 .sort()
                 .should.deep.equal(['file1.js', 'file2.js', 'package.json']);
 
-            exclude({ cwd: path.join(process.cwd(), 'test') })
+            exclude({ cwd: path.join(__dirname, 'fixtures') })
                 .globSync(cwd)
                 .sort()
                 .should.deep.equal([


packages/test-exclude/default-exclude.js
```diff
@@ -9,5 +9,6 @@ module.exports = [
     'test{,-*}.{js,cjs,mjs,ts}',
     '**/*{.,-}test.{js,cjs,mjs,ts}',
     '**/__tests__/**',
+    '**/fixture*{,/**}',
```
coreyfarrell: I don't object to ignoring `**/fixture/**` and `**/fixtures/**` but is `**/fixture*/**` excessive?  Maybe it would be better to use `'**/fixture{*,/**,s/**}`?  This way we ignore `fixture*` files but not `fixture*` directories?  In truth I'm still a bit concerned about ignoring `fixture*` files.

CC @bcoe @JaKXz interested in your feedback in this.  We're still open to breaking changes right now I just want to make sure this is a change that we want.  This would not break any of my own projects as I already have `fixtures/**` in the exclude list of my shared nyc config, I just worry because changes to the test-exclude defaults are a really big deal.

shinnn: [I personally name a directory `fixture-something` in my project](https://github.com/shinnn/eslint-config/tree/v7.0.0-0/test). Of course I could name it `fixtures/something` instead, but I wanted to reduce a frequency of `cd` while developing.

If you Istanbul maintainers think

> ignore `fixture*` files but not `fixture*` directories

is a more common default, I’ll update this PR.

@JaKXz Your 👍 means `**/fixture{*,/**,s/**}` looks better, right?

JaKXz: Yep!

shinnn: > In truth I'm still a bit concerned about ignoring `fixture*` files.

@coreyfarrell @JaKXz Can I ask a bit more details of the reason for this?

coreyfarrell: See comment from @bcoe about opting out of an exclude.  This is not easy to do with NYC, you either take the default excludes or replace them all.  The concern is that some people might use fixture files or directories for things that should be covered.

For my own projects I created/use [@cfware/nyc](https://github.com/cfware/cfware-nyc/) to generate configurations, in the case of that module `fixtures/**` and `helpers/**` is excluded by default.  For my one project which needs coverage for `fixtures/**` I added `'!fixtures/**'` to my exclude (negated exclude).  This is easier for me because my config generator appends the exclude list, so I just created nyc.config.js with `module.exports = require('@cfware/nyc').exclude('!fixtures/**').settings`.  For JSON based configs this is not so easy as only way to modify the exclude list is to completely replace it.

shinnn: I’m talking about why you are concerned about ignoring `fixture*` *files* but aren’t about `fixture*` *directories*. Or am I missing something in @bcoe’s comment?

bcoe: ideas I have to make this less annoying:

1. have a special exclude key that indicates all existing exclude rules should be included.
2. or, have a new `forced-include` key, or something along these lines, which takes precedence over include/exclude.
3. or, document the default rules in an easy to copy/paste line in README.md, with an example of how you'd copy them into your `package.json` config, with one dropped.

coreyfarrell: Maybe an option `excludeAppendDefaults: Boolean` (default false)?  This way someone could do:
```js
{
  "nyc": {
    "excludeAppendDefaults": true,
    "exclude": ["!fixtures/**"]
  }
}
```

packages/test-exclude/test/test-exclude.js
```diff
@@ -53,6 +53,28 @@ describe('testExclude', () => {
...
+        exclude()
+            .shouldInstrument('./fixture-some-suffix/index.js')
+            .should.equal(false);
+
+        exclude()
+            .shouldInstrument('./nested/fixtures/index.js')
+            .should.equal(false);
+
+        exclude()
+            .shouldInstrument('./fixtures-file.txt')
```
JaKXz: I think the default extension list would be excluding this, not the `fixture` name

packages/test-exclude/default-exclude.js
bcoe: I don't hate the idea of `excludeAppendDefaults`, as long as we have clearly defined behavior around order of operations; I assume `!fixtures` will be higher precedence than `fixtures/**`.

I think adding some notes into `nyc`'s README.md and `test-exclude`'s README.md will go a long way too.

I'd make adding `fixtures/` contingent on this above improvement.

coreyfarrell: Negated excludes always have higher priority.

packages/test-exclude/test/test-exclude.js
coreyfarrell: test-exclude doesn't actually have a default extension list, that's only in nyc.  Maybe this should change before the final semver-major release, move the default extension list from nyc to here?

In any case I'm leaning towards agreeing with @JaKXz that the tests here should just test against `.js` files.

packages/test-exclude/default-exclude.js
coreyfarrell: > I don't hate the idea of `excludeAppendDefaults`, as long as we have clearly defined behavior around order of operations; I assume `!fixtures` will be higher precedence than `fixtures/**`.
> 
> I think adding some notes into `nyc`'s README.md and `test-exclude`'s README.md will go a long way too.
> 
> I'd make adding `fixtures/` contingent on this above improvement.

I have to agree with @bcoe without the `excludeAppendDefaults` I don't wish to move forward on this.  Please note that `excludeAppendDefaults` would be semver-minor but adding `fixtures` to the default exclude would be semver-major.

Unfortunately I'm going to have to close this PR as we had to do some restructuring of code.  The list of default excludes now live at https://github.com/istanbuljs/schema/blob/master/default-exclude.js and the test-exclude module has been moved to it's own repository https://github.com/istanbuljs/test-exclude.  To proceed I think we would need a PR against test-exclude to implement `excludeAppendDefaults`, then we can add fixtures to the default excludes in `@istanbuljs/schema`.  The Window to add fixtures to the default exclude of nyc 15 is closing.

PR #465, opened 2019-09-09
feat: enable babel v8intrinsic plugin by default for v8 natives syntax

JounQin:


JaKXz:
I agree with you @coreyfarrell - this is easy enough to add in configuration for projects that use the node flags or extra Babel config.

JounQin:
How to config babel plugin for jest? That's why I send the PR.

coreyfarrell:
I don't think this PR would have any effect on jest.  jest performs it's own call to babel transform so the babel parser plugin would need to be set in `@jest/transform` (I don't see an option to do this).

JounQin:
@coreyfarrell @JaKXz It's not about transform, it's about coverage collection. My test cases passed, but the coverage collection failed because `istanbul-lib-instrument` can't recognize the valid syntax.

You can check https://github.com/rx-ts/babel-preset-proposal-typescript, remove the postinstall script first which does a trick for patching `istanbul-lib-instrument` inside `node_modules`, and simply run `yarn && yarn test`. It will throw an error, and I don't know how to prevent it in jest configuration.

JounQin:
@coreyfarrell @JaKXz Any idea please?

coreyfarrell:
I'm sorry I think this is a question for jest.  Our package accepts an option to configure the babel parser plugins, we cannot support zero-config for this situation.

PR #472, opened 2019-09-17
Use path.posix.relative to fix paths when generating html reports on Windows

### LINKED ISSUE #195: HTML StandardLinkMaper relativePath on windows
In `istanbul-reports\lib\html\index.js` there is a function that generates src paths for html references:
```js
relativePath: function (source, target) {
	var targetPath = this.getPath(target),
		sourcePath = path.dirname(this.getPath(source));
	return path.relative(sourcePath, targetPath);
}
```
This is used to generate html like this (on windows):
```html
<head>
    <title>Code coverage report for src</title>
    <meta charset="utf-8" />
    <link rel="stylesheet" href="..\prettify.css" />
    <link rel="stylesheet" href="..\base.css" />
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <style type='text/css'>
        .coverage-summary .sorter {
            background-image: url(..\sort-arrow-sprite.png);
        }
    </style>
</head>
```
Since the output from `path.relative` is platform specific and `\` is not standard this breaks things when built on windows. Eg `background-image: url(..\sort-arrow-sprite.png);` will fail on chrome and downstream parsing of html breaks in `gulp-inline-source`.

A simple solution would be:
```js
return path.relative(sourcePath, targetPath).replace(/\\/g, '/');
```
But I am no sure if this covers all the places paths are generated.

MayhemYDG:
Documentation:
https://nodejs.org/api/path.html#path_windows_vs_posix
https://nodejs.org/api/path.html#path_path_posix

Fixes #195

Tested this locally and it works great.

### DIFF
--- a/packages/istanbul-reports/lib/html/index.js
+++ b/packages/istanbul-reports/lib/html/index.js
@@ -94,7 +94,7 @@ const standardLinkMapper = {
     relativePath(source, target) {
         const targetPath = this.getPath(target);
         const sourcePath = path.dirname(this.getPath(source));
-        return path.relative(sourcePath, targetPath);
+        return path.posix.relative(sourcePath, targetPath);
     },
 
     assetPath(node, name) {


MayhemYDG:
@coreyfarrell Could you give this a quick look? It's a 1 work PR.

coreyfarrell:
Sorry for the delay I'm always squeamish about Windows patches since my only real access to Windows is via Travis-CI.  I see now that this only effects URL construction.  It's too bad that `require('url')` doesn't have a function for construction of relative URL's.

PR #489, opened 2019-10-08
feat: Convert to async API

coreyfarrell:
BREAKING CHANGE: MapStore#transformCoverage is now async and returns a
the coverage data only.  The `sourceFinder` method is now async and
provided directly on the `MapStore` instance.

---

CC @SimenB I know you use this API in jest, any objections?  We're not yet upgrading to `source-map@0.7.x` but this will make that upgrade possible without further changes to the public API.  Moving the `sourceFinder` method from the result of transformCoverage to the class itself is not strictly needed but just seems correct.

### DIFF
--- a/packages/istanbul-lib-source-maps/lib/map-store.js
+++ b/packages/istanbul-lib-source-maps/lib/map-store.js
@@ -6,10 +6,13 @@
 
 const path = require('path');
 const fs = require('fs');
+const { promisify } = require('util');
 const debug = require('debug')('istanbuljs');
-const SMC = require('source-map').SourceMapConsumer;
+const { SourceMapConsumer } = require('source-map');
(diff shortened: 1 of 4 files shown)


SimenB:
>CC @SimenB I know you use this API in jest, any objections?

Should be fine, the only things jest needs to be sync is source transformation and source maps for errors. Coverage calculation is perfectly fine async 👍

coreyfarrell:
What do you mean source maps for errors?  You mean production of source maps during the transformation or something else?

SimenB:
Nothing to do with istanbul, I just mentioned which parts related to sourcemaps Jest needs to have sync. (when running tests we might get errors that we wanna map through sourcemaps)

PR #492, opened 2019-10-23
feat: Add support for projectRoot option

bgalletbgallet:
Add the possibility to use --relative=true option in command line
This is useful if you run your test on a linux environment (docker, vm, server) and dev on a windows one.
If you want your coverage to work with your IDE, you need relative path instead of absolute ones.

Linked with pull request https://github.com/istanbuljs/nyc/pull/1212


packages/istanbul-reports/lib/lcovonly/index.js
```diff
@@ -9,6 +9,7 @@ class LcovOnlyReport extends ReportBase {
     constructor(opts) {
         super();
         this.file = opts.file || 'lcov.info';
+        this.relative = opts.relative || false;
```
coreyfarrell: Sorry for the delay responding, instead of having a `relative` boolean option lets have a `cwd` string option with a default `process.cwd()`.  I've discussed this with @bcoe and since we're in the process of making a semver-major release I think we can just make the lcov report always relative.  nyc will pass `cwd: this.cwd` to the report that way if someone set a `--cwd` different from `process.cwd()` the report will do the right thing.

bcoe:
Thanks for taking this work on @bgalletbgallet, exactly what I'm finding I need today.

packages/istanbul-reports/lib/lcovonly/index.js
bgalletbgallet: Hi @coreyfarrell it may have an issue on the version currently available on npm
If I add the --cwd=something option, the onDetail method of the lcovonly/index.js is not executed anymore.
It doesn't happen whith an other option, only with --cwd

coreyfarrell: I'm unsure how that happens, also not sure what you mean by "version currently available on npm".  If you're talking about 2.x that doesn't matter, this patch will only be applied to 3.x.

That said when I searched the istanbuljs repo for `cwd` I found the following in a couple reports:
```js
this.projectRoot = opts.projectRoot || process.cwd();
```

So probably best for lcov reports to do the same.

bgalletbgallet: I have the same issue on the current version, if I add a cwd option, onDetail is not executed.
Don't know if it's related, but it changes the directory where coverage files are generated

coreyfarrell: I understand now.  Using `nyc --cwd=somepath` does have effects beyond reporting.  So for the purpose of this PR I'm good, I still have a slight concern with adding a `projectRoot` option to nyc as I'm not sure if being able to have `cwd` different from `projectRoot` is useful or if it'll just create confusion.

In any case I'm happy with this change to istanbul-reports, thanks for the contribution.

PR #495, opened 2019-11-07
Fix html report due relative urls

ricardotejo:
- Fix missing images Windows because backslashes (missing path.posix in v2.2.5)
- Change folder link, removing '.index.html' because Chrome removes it.
- Fix breadcrumb folder url

The error is present in all versions, tested on istanbul-reports v2.2.5 and 3.0.0-alpha.1.

fix-detail

### DIFF
--- a/packages/istanbul-reports/lib/html/index.js
+++ b/packages/istanbul-reports/lib/html/index.js
@@ -94,7 +94,7 @@ const standardLinkMapper = {
     relativePath(source, target) {
         const targetPath = this.getPath(target);
         const sourcePath = path.dirname(this.getPath(source));
-        return path.posix.relative(sourcePath, targetPath);
+        return path.posix.relative(sourcePath, targetPath).replace(/index\.html$/, '');
     },
 
     assetPath(node, name) {
@@ -132,7 +132,7 @@ class HtmlReport extends ReportBase {
         const linkPath = nodePath.map(ancestor => {
             const target = this.linkMapper.relativePath(node, ancestor);
             const name = ancestor.getRelativeName() || 'All files';
-            return '<a href="' + target + '">' + name + '</a>';
+            return '<a href="' + (target == '' ? '/' + name : target) + '">' + name + '</a>';
         });
 
         linkPath.reverse();
--- a/packages/istanbul-reports/lib/html/index.js
+++ b/packages/istanbul-reports/lib/html/index.js
@@ -132,7 +132,7 @@ class HtmlReport extends ReportBase {
         const linkPath = nodePath.map(ancestor => {
             const target = this.linkMapper.relativePath(node, ancestor);
             const name = ancestor.getRelativeName() || 'All files';
-            return '<a href="' + (target == '' ? '/' + name : target) + '">' + name + '</a>';
+            return '<a href="' + target + '">' + name + '</a>';
         });
 
         linkPath.reverse();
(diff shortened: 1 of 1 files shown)


packages/istanbul-reports/lib/html/index.js
```diff
@@ -94,7 +94,7 @@ const standardLinkMapper = {
     relativePath(source, target) {
         const targetPath = this.getPath(target);
         const sourcePath = path.dirname(this.getPath(source));
-        return path.posix.relative(sourcePath, targetPath);
+        return path.posix.relative(sourcePath, targetPath).replace(/index\.html$/, '');
```
coreyfarrell: This change cannot be merged, many people (myself included) view the istanbul reports via filesystem.  So I'll run `nyc report --report html && firefox coverage/index.html`.  Removing the `index.html` breaks this report for viewers that are not viewing through a web server while also requiring that the web server be configured for `index.html` as directory index.

packages/istanbul-reports/lib/html/index.js
```diff
@@ -132,7 +132,7 @@ class HtmlReport extends ReportBase {
         const linkPath = nodePath.map(ancestor => {
             const target = this.linkMapper.relativePath(node, ancestor);
             const name = ancestor.getRelativeName() || 'All files';
-            return '<a href="' + target + '">' + name + '</a>';
+            return '<a href="' + (target == '' ? '/' + name : target) + '">' + name + '</a>';
```
coreyfarrell: I'm unclear on this change, if target is empty then we would use `name`?  This means we could end up with `<a href="All files">All files</a>`.

packages/istanbul-reports/lib/html/index.js
ricardotejo: I've made a new change because you are right, my change was not working in browser directly. So now the new approach is working in browser and server. Please review it.

packages/istanbul-reports/lib/html/index.js
ricardotejo: This is not needed if my previous change is accepted.

coreyfarrell:
I'm not comfortable with this patch, in particular the use of string manipulation on URL's paths is something I'm not keen on.  Before I can evaluate any further I need to know if the issue you've experienced exists in `istanbul-reports@3.0.0-alpha.3`.  You can test this by `npm install nyc@next istanbul-reports@next`.  Once that is done please run `npm ls istanbul-reports` to verify that only 3.0.0-alpha.3 is installed.  In particular if 3.0.0-alpha.3 produces HTML which fails to render correctly I need to see examples of broken HTML code produced by that version.

I appreciate you contributing a fix and I'm not trying to be dismissive, this is just a critical feature so I need to be really careful (our testing of the generated HTML report is lacking).

ricardotejo:
I've tested 3.0.0-alpha.3 version isolated in a clean project and got the same problem when it runs within a web server, after testing with more web-servers I found that the problem was in the this side, the url rewrite module of these servers work different.

In Nginx works fine when we use /index.html but fails with a rewrite. 
http-server fails in v0.11.1 but works fine in 0.9.0 /index.html
serve removes /index.html and failed on directories.

Maybe this problem would be addressed with a "notice" in the documentation to use "/index.html" urls.

Thank you the time you invested reviewing my commits.

coreyfarrell:
You should be able to configure a web server to use `index.html` as the directory index and the page should still work.  So in nginx `index index.html;` should allow you to navigate to the istanbul page without issue, the the pages themselves will still link to `index.html`.

PR #502, opened 2019-11-18
fix:handlebars arbitrary code execution

henriquecaraujo:
Update Handlebars dependency to fix the arbitrary code execution.

> Versions of handlebars prior to 4.5.2 are vulnerable to Arbitrary Code Execution The package's lookup helper fails to properly validate templates, allowing attackers to submit templates that execute arbitrary JavaScript in the system. It can be used to run arbitrary code in a server processing Handlebars templates or on a victim's browser (effectively serving as Cross-Site Scripting).
([https://www.npmjs.com/advisories/1316](https://www.npmjs.com/advisories/1316))

### DIFF
--- a/packages/istanbul-reports/package.json
+++ b/packages/istanbul-reports/package.json
@@ -1,60 +1,60 @@
 {
-    "name": "istanbul-reports",
-    "version": "3.0.0-alpha.3",
-    "description": "istanbul reports",
-    "author": "Krishnan Anantheswaran <kananthmail-github@yahoo.com>",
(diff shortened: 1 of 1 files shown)


coreyfarrell:
This is not a bug in istanbul-reports, the problem is that you have a package-lock.json or yarn.lock which is blocking the in-range update.  If you delete the lockfile and reinstall your `npm audit` will be empty.  The other option is to use `npm audit fix` as instructed by npm.

Note that I'm actually going to be removing handlebars entirely, see #503.

PR #511, opened 2019-12-11
feat(text): Coalesce ranges of missing lines

piranna:
This pull-request group ranges of lines showind only the first and last ones (separated by an hyphen) so some chars can be earned and more missing lines can be shown. This behaviour is the same of [coverage.py](https://coverage.readthedocs.io).


coreyfarrell:
`Array#flatMap` is a node.js 12 feature so we cannot use it yet.  For node.js 8+ I generally use code such as `[].concat(...resultOfMap)`.

In addition since this did not cause any of the existing tests to change I have to request that you add a test to cover this new feature.

piranna:
> `Array#flatMap` is a node.js 12 feature so we cannot use it yet. For node.js 8+ I generally use code such as `[].concat(...resultOfMap)`.

The curious thing is that the Windows builds for Node.js 8 and 10 has passed... A bug in Travis CI? Anyway, good trick, it works as expected :-)


> In addition since this did not cause any of the existing tests to change I have to request that you add a test to cover this new feature.

I was a bit surprised too, but it made sense since the actual tests didn't have any fixture with more than two consecutive missing lines, just only the "too many missing lines" one, that have them at the very begining of the file, so they were not being shown in the report. I've created a new fixture/test based on it and modified the lines by hand to have several consecutive ranges, and added the new test :-)

piranna:
Tests are now passing with Node.js 8 and 10 too.

coreyfarrell:
> The curious thing is that the Windows builds for Node.js 8 and 10 has passed... A bug in Travis CI? Anyway, good trick, it works as expected :-)

The TextReporter tests are skipped on Windows (for some reason I'm unsure without digging).  In any case this patch has nothing Windows specific so this does not hold us back.  Thanks for the contribution.

piranna:
> The TextReporter tests are skipped on Windows (for some reason I'm unsure without digging).

Maybe it's related to fetch the console size or something similar. It could be interesting to try them, but don't have a Windows machine at hand at this moment.


> In any case this patch has nothing Windows specific so this does not hold us back. Thanks for the contribution.

Great, thank you! :-D

coreyfarrell:
I also do not have a Windows machine and using travis-ci to do this kind of testing is cumbersome.  Also of note the nyc tests do run on Windows and snapshot testing of text report is performed so it does get coverage on Windows by other means.

piranna:
> Also of note the nyc tests do run on Windows and snapshot testing of text report is performed so it does get coverage on Windows by other means.

Good to know :-)

bcoe:
@piranna just noticed this feature, I love it 👍 https://github.com/bcoe/c8/pull/179/files#diff-123ced42a67a2ddad8bff758c2b8604a

piranna:
> @piranna just noticed this feature, I love it  [bcoe/c8/pull/179/files#diff-123ced42a67a2ddad8bff758c2b8604a](https://github.com/bcoe/c8/pull/179/files#diff-123ced42a67a2ddad8bff758c2b8604a)

Thank you, I'm glad you liked it! :-) Your `c8` project looks cool too :-D

PR #515, opened 2019-12-28
fix: Update to support nyc 15.

coreyfarrell:
BREAKING CHANGE: This configuration is no longer valid for nyc 14 and
below.

---

@JaKXz do you agree that this is a breaking change?  Sorry I didn't catch this before releasing 1.0.0.

### DIFF
--- a/packages/nyc-config-typescript/index.js
+++ b/packages/nyc-config-typescript/index.js
@@ -1,16 +1,8 @@
 'use strict';
 
+const { parserPlugins } = require('@istanbuljs/schema').defaults.nyc;
+
 module.exports = {
     cache: false,
-    extension: ['.ts', '.tsx'],
-    exclude: [
-        '**/*.d.ts',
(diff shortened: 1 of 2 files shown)


coreyfarrell:
Additional question, do you think this should have a peer dep on nyc >=15 or should that be done through documentation only?

coreyfarrell:
> I think v1.0.0 should have the peer dep on nyc >= 15 👍🏽
> 
> This looks great!

I agree that 1.0.0 should have the peer dep, unfortunately 1.0.0 was already published 8 days ago.  So the question now do I publish this as 1.1.0 (treat 1.0.0 as a broken release) or do another semver-major to 2.0.0?

JaKXz:
Oh - then this should be a patch just so that people get the warning when they are installing v ^1.0.0

coreyfarrell:
@JaKXz thanks for the review, 1.0.1 is now published.

PR #528, opened 2020-02-12
fix wrong computed name in cobertura when using v8 coverage

jehy:
This PR fixes https://github.com/istanbuljs/istanbuljs/issues/527 -  when using v8 coverage, you sometimes get `<computed>` in message name, and cobertura can't process generated file with message like this:

```
11:09:42  FATAL: Unable to parse builds/8/coverage.xml
11:09:42  hudson.util.IOException2: Cannot parse coverage results
11:09:42  	at hudson.plugins.cobertura.CoberturaCoverageParser.parse(CoberturaCoverageParser.java:84)
11:09:42  	at hudson.plugins.cobertura.CoberturaCoverageParser.parse(CoberturaCoverageParser.java:52)
11:09:42  	at hudson.plugins.cobertura.CoberturaPublisher.perform(CoberturaPublisher.java:594)
11:09:42  	at org.jenkinsci.plugins.workflow.steps.CoreStep$Execution.run(CoreStep.java:80)
11:09:42  	at org.jenkinsci.plugins.workflow.steps.CoreStep$Execution.run(CoreStep.java:67)
11:09:42  	at org.jenkinsci.plugins.workflow.steps.SynchronousNonBlockingStepExecution.lambda$start$0(SynchronousNonBlockingStepExecution.java:47)
11:09:42  	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
11:09:42  	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
11:09:42  	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
11:09:42  	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
11:09:42  	at java.lang.Thread.run(Thread.java:748)
11:09:42  Caused by: org.xml.sax.SAXParseException; lineNumber: 7441; columnNumber: 31; The value of attribute "name" associated with an element type "method" must not contain the '<' character.
```

PR #534, opened 2020-03-08
fix: cobertura should escape invalid characters

### LINKED ISSUE #527: Cobertura reporter makes non valid method when using v8 coverage
If you use Node 12 or 13 and v8 coverage provider, istanbul generates wrong method names, including non valid `<` and `>` characters, like this:

```
<method name="map.<computed>.f" hits="1" signature="()V">

```
Bug can be reproduced with jest and with c8:
facebook/jest#9561
bcoe/c8#149

I can try to provide a minimal example, but it seems like a simple fix would be to replace `<` and `>` with `&lt;` and `&gt;`.

bcoe:
see: https://github.com/bcoe/c8/issues/149

Fixes #527
Closes #528

### DIFF
--- a/packages/istanbul-reports/lib/cobertura/index.js
+++ b/packages/istanbul-reports/lib/cobertura/index.js
@@ -4,6 +4,7 @@
  Copyrights licensed under the New BSD License. See the accompanying LICENSE file for terms.
  */
 const path = require('path');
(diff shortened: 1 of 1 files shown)


jehy:
What about a new release with this fix?

PR #573, opened 2020-06-17
lcovonly - onDetail - allow `meta` to be undefined

### LINKED ISSUE #572: TypeError: Cannot read property 'loc' of undefined
This line sometimes throws an error:

https://github.com/istanbuljs/istanbuljs/blob/1b52fe750d1f800c34dbff168614c0c73bd76026/packages/istanbul-reports/lib/lcovonly/index.js#L53

```
    TypeError: Cannot read property 'loc' of undefined
        at /node_modules/nyc/node_modules/istanbul-reports/lib/lcovonly/index.js:53:35
```

This is causing issues when using downstream utilities such as:

- @cypress/code-coverage https://github.com/cypress-io/code-coverage/issues/216#issuecomment-620618012
- jest https://github.com/facebook/jest/issues/9659

penx:
I'm not 100% on this as I'm not familiar with istanbul's inner workings, but this seems like a sensible check.

Fixes #572

### DIFF
--- a/packages/istanbul-reports/lib/lcovonly/index.js
+++ b/packages/istanbul-reports/lib/lcovonly/index.js
@@ -50,6 +50,9 @@ class LcovOnlyReport extends ReportBase {
 
         Object.entries(branches).forEach(([key, branchArray]) => {
             const meta = branchMap[key];
+            if (!meta) {
+                return;
+            }
             const { line } = meta.loc.start;
             branchArray.forEach((b, i) => {
                 writer.println('BRDA:' + [line, key, i, b].join(','));


bitcloud:
I just had a quick look at the code and I guess just ignoring branches id's that you can't resolve through the branchMap seems to ignore the real issue here.

Check out this test data:
https://github.com/istanbuljs/istanbuljs/blob/master/packages/istanbul-lib-coverage/test/file.test.js#L153

It comes down to those two parts:
```js
            branchMap: {
                1: {
                    type: 'if',
                    line: 2,
                    locations: [loc(2, 1, 2, 20), loc(2, 50, 2, 100)]
                }
            },
            b: {
                1: [0, 0]
            }
```

`BranchMap` contains the position of your branch head and its branch locations while `b` just references to an element of branchMap and counts the hits of the single branch locations.

This means by ignoring the hitcounts of branches because you are not finding an entry in the map, you are actually ignoring the whole branch altogether then.

As you can see here, the entry in the map should actually be the basis of the hit counts:
https://github.com/istanbuljs/istanbuljs/blob/master/packages/istanbul-lib-source-maps/lib/mapped.js#L71

So it seems that there is a bigger issue if there is a hit count without a map entry.

bitcloud:
It would be great if you could provide a test case which triggers this issue. That would help to reproduce the issue and figure out what is really happening.

coreyfarrell:
We've had many reports of this kind, unfortunately this PR simply covers up a legitimate error in the data.  meta is required, if not present the data is broken.  This is likely caused by attempting to merge incompatible code (for example testing `src/file.js` once without transpiling, once with ES5 transpiling).  I'm not sure what the solution to this issue is but ignoring it is not the correct fix.  nyc 15 fixes many issues of this kind with changes to how it handles `--all=true` but other issues remain, largely caused by testing with inconsistent transpile settings.

PR #587, opened 2020-11-03
sourceFinder support async

songjiachao:
I want to read file by gitlab api , but sourceFinder can't support async.

### DIFF
--- a/packages/istanbul-lib-report/lib/context.js
+++ b/packages/istanbul-lib-report/lib/context.js
@@ -67,8 +67,8 @@ class Context {
      * @param {String} filePath the file path as found in a file coverage object
      * @returns {String} the source code
      */
-    getSource(filePath) {
-        return this.sourceFinder(filePath);
+    async getSource(filePath) {
+        return await this.sourceFinder(filePath);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


coreyfarrell:
This would be a massive breaking change, I'm not sure if it's even possible.  Changing functions from sync to async cause ripples across the code.  This would require every function which calls `onDetail` to become async, every function which calls that, etc.  During the pre-release of current major release I made sourceFinder async but then reverted it due to issues with reporting.  I don't remember if this was simply that I didn't have time to deal with switching all of reporting to async or if there was a specific blocker.

I'm not saying this is off the table but it would need to be proven that this can work.  If this were to happen it would be part of a major release, I do not have an ETA for that.  We would also need to check with jest developers (CC @SimenB) to ensure this would not break them - they are a major user of istanbul-reports and I've seen some cases were async functions simply cannot be used.

SimenB:
Jest's reporting is async, so it would not be a big issue for us to migrate to an async API 👍 Instrumentation would need to stay sync (that's a Babel plugin, so probably not affected here), but reporting can be async without issue from Jest's perspective

coreyfarrell:
@SimenB No possibility of instrumentation becoming async as long as `require` API exists.  Thanks for the feedback about the reporting API.

PR #591, opened 2020-12-30
fix: reverse tabnabbing vulnerability in URLs

AnirudhMergu:
Tabnabbing - "it's the capacity to act on parent page's content or location from a newly opened page via the back link exposed by the opener javascript object instance."

This vulnerability usually manifests when either

The "target" attribute is used to specify the target location in an anchor <a> tag to open 3rd party URL/resource(s) without including the attribute rel="noopener,noreferrer " in the anchor <a> tag

noopener - When noopener is set, the child window opened will not have access to the parent window via "Window.opener" and at the same time, the parent window will not have access to the child window. "Note that when noopener is used, nonempty target names other than _top, _self, and _parent are all treated like _blank in terms of deciding whether to open a new window/tab."

noreferrer - When noreferrer is set, the child window opened will not have access to parent URL. "In addition, setting this feature also automatically sets noopener."


bcoe (approved):

bcoe (approved):

bcoe (approved):

bcoe:
@AnirudhMergu thank you for the contribution \o/

PR #608, opened 2021-05-13
feat: add line colorMetric option

opichals:
Makes the report highlight the lines accoording to the chosen metric.

This helps in creating the appropriate perception of the generated output (especially the html version) based on the metric the users decide to base to evaluate their coverage data upon.


bcoe:
@opichals I'm not 100% following this feature, would you be able to provide screen shots of how you would use the functionality in an HTML report?

Issue #322, opened 2019-03-06
TypeError: Cannot read property 'locations' of undefined


bennyn:
I created a coverage report using the `nyc karma start` command. It generated a HTML report but when I click on a class in that report I see the following error:

> Cannot read property 'locations' of undefined
TypeError: Cannot read property 'locations' of undefined
    at D:\dev\projects\temp\karma-test\node_modules\nyc\node_modules\istanbul-reports\lib\html\annotator.js:133:48
    at Array.forEach ()
    at annotateBranches (D:\dev\projects\temp\karma-test\node_modules\nyc\node_modules\istanbul-reports\lib\html\annotator.js:128:30)
    at Object.annotateSourceCode (D:\dev\projects\temp\karma-test\node_modules\nyc\node_modules\istanbul-reports\lib\html\annotator.js:238:9)
    at HtmlReport.onDetail (D:\dev\projects\temp\karma-test\node_modules\nyc\node_modules\istanbul-reports\lib\html\index.js:269:23)
    at Visitor.(anonymous function) [as onDetail] (D:\dev\projects\temp\karma-test\node_modules\nyc\node_modules\istanbul-lib-report\lib\tree.js:34:30)
    at ReportNode.Node.visit (D:\dev\projects\temp\karma-test\node_modules\nyc\node_modules\istanbul-lib-report\lib\tree.js:121:17)
    at D:\dev\projects\temp\karma-test\node_modules\nyc\node_modules\istanbul-lib-report\lib\tree.js:114:23
    at Array.forEach ()
    at visitChildren (D:\dev\projects\temp\karma-test\node_modules\nyc\node_modules\istanbul-lib-report\lib\tree.js:113:32)

Scenario can be replayed with the follwoing repository:
- https://github.com/bennyn/typescript-jasmine-karma-nyc/commit/3fd49eb9f5c99518e32f7cfff37f3414305d586c

coreyfarrell:
It looks like karma-coverage depends on the deprecated istanbul module.  I suspect this is why it is generating coverage files that are incompatible with nyc (lacking fields required by the HTML report).  See karma-runner/karma-coverage#375 for the ticket about migrating away from the `istanbul` module.

zhaochy1990:
hello @coreyfarrell , I also meet this error in my project, is there any update for this issue?

I'm writing Typescript codes, and only run these codes on Node.js environment.

I'm using
nyc: 14.1.1
node.js: 10.15.3
[...]

PR #617, opened 2021-07-03
fix: merge branch/statement/functionMap's together when merging two coverage reports

JustinChristensen:
@bcoe @coreyfarrell 

This is an attempt to fix the file coverage merge problems described in this issue, and a few other open issues.
[https://github.com/istanbuljs/istanbuljs/issues/322#issuecomment-873349144](https://github.com/istanbuljs/istanbuljs/issues/322#issuecomment-873349144)

This hasn't been extensively tested, and I'm mainly just spitballing on what merge should look like. 

This is a breaking change.

I should also mention that I don't know all of the invariants that should be assumed to hold prior to merging two file coverages. I'm hoping that someone with more background on this code might be able to weigh in.


bcoe:
@JustinChristensen  thanks for digging into this... because of how finicky this problem is, I would like to be very careful, and would want to test with `v8-to-istanbul` as well.

Would it be possible to add tests that represent the scenario you outline in your comment?

JustinChristensen:
@bcoe 

I would say that it's maybe a little early to be talking about tests for this code just yet. I was hoping that someone with more background on this codebase might be able to first review it and do a first blush sanity check, and give some idea of whether or not my assessment of the problem was correct and whether this seems like it's on the right track in being a fix for it (and whether there might be other places that have a similar problem that also need fixing). 

This is finicky, but it's also trivial to verify that a mismatch in the number of properties between an item's "stats" and "meta" would lead code like https://github.com/istanbuljs/istanbuljs/blob/f6bb0b4f3e7596a4e2dcaa1d0695afe1f97e6eee/packages/istanbul-reports/lib/html/annotator.js#L127 to error. The question is whether this is the only such case where these pairs of properties might get out of sync.

Let me know how you want to proceed.

JustinChristensen:
@bcoe Also, it might not give you as much confidence as an actual unit test, but I did verify this with the FileCoverage objects I mentioned in that comment, and more broadly with all of the coverage objects in my projects' test suite, and this seemed to work just fine.

JustinChristensen:
Some potential areas of concern:
1. Is the start and end range on the `loc` property of branches and functions, and on statements themselves, enough to uniquely identify these objects while taking the union of the two sets?
2. Will there always be the `loc` property or both `start` and `end`? Does this codebase enforce that invariant?
3. What's the point of `all`? 

Also, I just tried running the tests, and I see that the eslint configuration for this project doesn't like my style. Fixing that now.

JustinChristensen:
@bcoe 

Heh, the way the unit tests are written indicate that the properties in `brancMap` won't have a `loc` property, and yet the branchMap properties from v8-to-istanbul do (as you can see in the `FileCoverage` objects I shared in that gist in the comment).

So I think I have my answer for #2 in the above list.

JustinChristensen:
I added `all` back in, because there's a test that expected it. 

Another test was expecting "map" properties where the keys started at 1 instead of 0 for some arbitrary reason. Instead of making the code keep track of those keys during the merge, I opted just to shift the keys over in the test, as I can't imagine why it would need to be that way.

bcoe:
@JustinChristensen perhaps we could ask the folks in #322 to provide a repo with a minimum reproduction of the merging issue? This would make it easier for me to understand the problem, also I'd like to test against c8 as well.

JustinChristensen:
@bcoe Oh, I see, you were asking for tests to help explain the issue. I thought we had already sufficiently done that in prose, but if you're looking for some additional assistance I can work on putting together an example for you.

JustinChristensen:
@bcoe 

Alternatively you can clone https://github.com/JustinChristensen/react-sample-app, and run `yarn start` and then in another tab `yarn e2e`, and then open `./coverage/src/components/EmployeeFields.jsx.html' in your browser.

bcoe:
@JustinChristensen I'm unlikely to have cycles to work on this in the immediate term. As long as we're testing in linked against `c8` and `v8-to-istanbul` as well, I'd be pretty confident with changes.

bcoe:
@JustinChristensen, @linliTian apologies, unfortunately no timeline for merging this, I've had very little time to dedicate to open source recently, and this change is finicky enough that I'll need to dig into it deeply -- _I'm sure that Justin has put great work in, but I need to do some manual testing in a variety of environments, since this is a non-trivial change to one of the more finicky parts of the codebase._

packages/istanbul-lib-coverage/lib/file-coverage.js
```diff
@@ -166,24 +207,37 @@ class FileCoverage {
...
+            other.s,
+            other.statementMap
+        );
+        this.data.s = hits;
+        this.data.statementMap = map;
+
+        const keyFromLocProp = x => keyFromLoc(x.loc);
+        const keyFromLocationsProp = x => keyFromLoc(x.locations[0]);
+
+        [hits, map] = mergeProp(
```
bcoe, line 222: One slight concern I have about this change is the number of additional operations that will be performed on each coverage report as we iterate over all these branch coverage operations.

Maybe there's no way around it, if this is necessary for correct merging behavior, but it would be good to find a large codebase to test against.

JustinChristensen: @bcoe The previous incarnation looped through each of these properties, including the branch hit counters, for just the "other" object. This new algorithm also needs to iterate through the properties for "this" object too. So in the worst case I would expect this to be potentially twice as slow, but that's a low price to pay for correctness.

JustinChristensen:
@bcoe Also, I yeah I figured you were busy. Are you the only maintainer for this project? I figured istanbul would be a little more active with how important of a role it places in the JS ecosystem.

packages/istanbul-lib-coverage/lib/file-coverage.js
bcoe: @JustinChristensen I'm pretty sure on large codebases using `c8`, it won't ultimately merge many coverage files because the merge happens at prior to processing when the output is in `v8` format.

I suppose the concern is a large codebase using `tap` and `nyc`, the `npm` CLI might be a good repo to test with -- if it works with npm, I'd be pretty confident we don't have too huge of a performance hit.

packages/istanbul-lib-coverage/test/file.test.js
```diff
@@ -153,36 +153,36 @@ describe('base coverage', () => {
         const template = new FileCoverage({
             path: '/path/to/file',
             statementMap: {
-                1: loc(1, 1, 1, 100),
-                2: loc(2, 1, 2, 50),
-                3: loc(2, 51, 2, 100),
-                4: loc(2, 101, 3, 100)
+                0: loc(1, 1, 1, 100),
```
bcoe, line 156: It seems a bit odd to me that this new logic would have switched tests to being 0 offset vs., 1 offset.

Is this expected?

bcoe:
@JustinChristensen I put some work in this weekend to make [releases of `istanbulsjs/istanbuljs` fully automated](https://github.com/istanbuljs/istanbuljs/pull/625), so that it's easier to take patches like this, and to iterate quickly if we do find we bump into any hiccups, e.g., misunderstood behavior slightly and introduce a bug.

This gives me more confidence to work on landing a patch like this 👍 I appreciate having someone dive into this codebase, and I hope how long I've taken to land this PR doesn't discourage you from further contributions (_help is very much needed and wanted._)

packages/istanbul-lib-coverage/test/file.test.js
JustinChristensen: @bcoe 

The previous incarnation of merge [preserved the object keys while zipping up the hit counters](https://github.com/istanbuljs/istanbuljs/blob/f6bb0b4f3e7596a4e2dcaa1d0695afe1f97e6eee/packages/istanbul-lib-coverage/lib/file-coverage.js#L169). [As those keys are just ascending integers](https://gist.github.com/JustinChristensen/aa256e0187dc638f7db08a1391136861#file-first-coverage-json-L626), and not a kind of identifier, it didn't seem particularly important to preserve them. Unless that behavior is documented somewhere, I suspect that the test was written with that unimportant implementation detail in mind, and not to validate an API's guarantees.

JustinChristensen: Actually, I probably should check the spec on that.

JustinChristensen: https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-ordinaryownpropertykeys

Yup.

bcoe: sounds good 👍

bcoe (approved):

bcoe:
@JustinChristensen thank you for the contribution, next step is that I'd like to test on a few large real world projects before we release.

JustinChristensen:
@bcoe Sounds good. I don't maintain any large open source projects to test this on, and so you'll have to let me know how the testing goes. I'm happy to make any fixes that are needed based on your findings.

dementors-magnus:
@bcoe Would it be possible to make an "@next" release to make it easier to test? We have a private repo with the same error as in #572 that I would like to check if this solves the issue.

bcoe:
@JustinChristensen thank you very much for the contribution 👏 

I've tested so far on `c8`, `yargs`, and one of the libraries I work on at Google, behavior seems to be stable so far. Excited to see if this is an improvement for people.

@dementors-magnus let us know if this does the trick for you/

dementors-magnus:
@bcoe Tested and from our perspective #572 is resolved with this fix! A big thank you to @JustinChristensen!

PR #619, opened 2021-07-29
WIP: feat: add platform-conditional ignore hints

isaacs:
This adds the ability to create ignore hints that are only effective
on specific platforms, so that we can ignore some test branches that
are only relevant to specific platforms by using a hint like:

/* istanbul ignore next platform!=win32 */
/* istanbul ignore else platform=linux */
and so on.

No tests yet.

### DIFF
--- a/packages/istanbul-lib-instrument/src/visitor.js
+++ b/packages/istanbul-lib-instrument/src/visitor.js
@@ -5,9 +5,9 @@ import { SourceCoverage } from './source-coverage';
 import { SHA, MAGIC_KEY, MAGIC_VALUE } from './constants';
 
 // pattern for istanbul to ignore a section
-const COMMENT_RE = /^\s*istanbul\s+ignore\s+(if|else|next)(?=\W|$)/;
+const COMMENT_RE = /^\s*istanbul\s+ignore\s+(if|else|next)(?:\s+platform(!?)=([a-z0-9]))?(?=\W|$)/;
 // pattern for istanbul to ignore the whole file
-const COMMENT_FILE_RE = /^\s*istanbul\s+ignore\s+(file)(?=\W|$)/;
+const COMMENT_FILE_RE = /^\s*istanbul\s+ignore\s+(file)(?:\s+platform(!?)=([a-z0-9]))?(?=\W|$)/;
 // source map URL pattern
 const SOURCE_MAP_RE = /[#@]\s*sourceMappingURL=(.*)\s*$/m;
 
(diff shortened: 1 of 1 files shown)


isaacs:
I don't know if this is something that would be acceptable, or if it should be somehow abstracted to look at `navigator.userAgent` in web browsers?  The syntax is an attractive bikeshed, I tried to keep it as simple as possible to meet the needs I have.

ljharb:
Maybe instead of hardcoding “platform”, conditional ignores could be generic - ie, you put whatever condition name you want in the comment, and the config/command-line invocation would determine whether that condition name is undefined, true, or false, and it’d be a noop when undefined?

isaacs:
To avoid a breaking change, maybe the syntax could be:

```js
/* istanbul ignore-conditional (file|if|else|next) (condition(s)...) comment */
```

So the first item that isn't a conditional means the rest is just comment, and users can always use `--` to separate out the comment section.

Ignore hint will be valid if ALL conditions are met.  (To OR the conditions, use multiple hints.)

Conditions:

- `key=value` true if the key is exactly equal to the value specified
- `key!=value` true if the key is not exactly equal to the value specified
- `key~=value` true if the key contains the value
- `key!~=value` true if the key does not contain the value

Some thoughts on extending this to more generic/unbounded set of keys and values might look like this.

Keys and values may be double-quoted, so `env.FOO="this is env foo"` is valid.  Quotes in quoted values may be escaped, so `env.QUOTED="this \" contains \" quotes"` is valid as well.

Keys available:

- `platform` compared with process.platform, or `""` if `process` or `process.platform` not available.
- `arch` compared with process.arch, or `""` if `process` or `process.arch` not available.
- `env.${KEY}` compared with relevant field in `process.env`, or "" if not available.
- `argv[n]` compared with `process.argv[n]`, or `""` if not available.

I'm not sure how to go about doing ignores for browsers.  Would that entail some kind of run-time handling?  Or is instanbul itself sometimes run within a browser context?

The downside of having open-ended values (instead of process.platform or process.arch, which have a small number of possible values) is that the parsing gets more involved, since environment variables and arguments may contain spaces, quotes, and escape characters.

ljharb:
env values can be provided pretty easily by bundlers to browsers, so that might be sufficient?

bcoe:
Just started looking at this repo a little bit again 👍 will make an effort to give this review.

isaacs:
> Just started looking at this repo a little bit again 👍 will make an effort to give this review.

@bcoe awesome!  Feel free to ping me on the various slacks if you have any feedback on it.

PR #627, opened 2021-09-14
fix: remove unnecessary peer dependencies

DylanPiercey:
Currently `ts-node` and `source-map-support` are listed as peer dependencies (which now forces them to be installed in npm7).
I don't believe these are strictly necessary, there are many require hooks for typescript (eg esbuild &  swc) and node now supports source maps natively.


bcoe (approved):

bcoe:
@DylanPiercey thank you for the contribution, sorry for taking so long to land.

PR #629, opened 2021-09-24
Feature: Evaluate Logical Truthiness

adrian-burlacu-software:
Istanbul is a great tool that I intend on using :D Yet, it fails to report the simple matter of logical truthiness of logical expressions. 

For example, expressions such as:

x && y 

This does report coverage of X and not Y even when X is false. But I want to know if these branches are ever able to become true. 

Another simple example:

x || y

When Y is true, the coverage of X is reported even though it never becomes true.

This change handles these cases in a new counter that piggibacks on the branches count, called branchesTrue(bT).

Some caveats:
- Minor version bump, istanbul-lib-instrument depends on the new istanbul-lib-coverage.
- Zero configuration, this always runs.
- Evaluates the expression, so the code should not mutate state in a logical expression.
- Only for logical expressions, otherwise the instrumented code needs to save the state to a temp variable(memoization), and re-evaluate its truthiness.

Thanks to the team working on this beloved library, I hope you find the time and willingness to merge this tested and necessary pull request into the master branch.


bcoe:
> This was clearly in the caveats. Don't expect to update state and evaluate logical expressions at the same time if you expect to allow for straightforward analysis.

Apologies missed this.

> Yes, this is evaluating the expression once for counting, once for the program. Ok,

I think I was misunderstanding, reading your code you represent the statement as this right:

`a() && b`  ➡️ `(a() && (cov.bT++, null)) && (b && (cov.bT++, null))`

`a() || b`  ➡️ `(a() && (cov.bT++, null) || (b && (cov.bT++, null))`

This would only evaluate `a()` once right?

bcoe:
> In fact, I wrote Fast-Fuzz, a coverage-guided fuzzing tool based on this commit, and it's on V2 already.

@adrian-burlacu-software that's quite cool, I'd love to play with try this out on some of my libraries.

adrian-burlacu-software:
@bcoe 
[Fast-Fuzz](https://github.com/WeWatchWall/fast-fuzz) has been updated with this fix and all the tests pass. Feel free to check out the new features on my fuzzer, it's now been tested more realistically.

Thanks again for helping me with all the suggestions :)

PR #662, opened 2022-01-02
Fixed an error when inputSourceMap is not a plain object

marekdedic:
This PR is a continuation of #576 - thanks to @JLHwung.


packages/istanbul-lib-instrument/src/visitor.js
```diff
@@ -697,9 +697,18 @@ function programVisitor(types, sourceFilePath = 'unknown.js', opts = {}) {
                 .update(JSON.stringify(coverageData))
                 .digest('hex');
             coverageData.hash = hash;
+            const inputSourceMap = coverageData.inputSourceMap;
+            if (inputSourceMap && Object.getPrototypeOf(inputSourceMap) !== Object.prototype) {
+              console.warn("A non-plain object was passed to istanbul-lib-instrument.")
```
marekdedic, line 702: @bcoe Feel free to suggest edits to the message...

packages/istanbul-lib-instrument/src/visitor.js
```diff
@@ -697,9 +697,18 @@ function programVisitor(types, sourceFilePath = 'unknown.js', opts = {}) {
...
+              const {...inputSourceMapPlain} = inputSourceMap;
+              coverageData.inputSourceMap = inputSourceMapPlain;
+            }
             const coverageNode = T.valueToNode(coverageData);
             delete coverageData[MAGIC_KEY];
             delete coverageData.hash;
+            if (inputSourceMap) {
+              coverageData.inputSourceMap = inputSourceMap;
+            }
```
bcoe, line 711:
```suggestion
            const inputSourceMap = coverageData.inputSourceMap;
            if (inputSourceMap && Object.getPrototypeOf(inputSourceMap) !== Object.prototype) {
              console.warn("A non-plain object was passed to istanbul-lib-instrument.")
              coverageData.inputSourceMap = {...inputSourceMap}
            }
            const coverageNode = T.valueToNode(coverageData);
            delete coverageData[MAGIC_KEY];
            delete coverageData.hash;
```

If I'm understanding the problem properly, could the implementation be simplified to this ☝️ 

Also, it would be nice to add a test, I think you could just add an extra test case here:

https://github.com/istanbuljs/istanbuljs/blob/c1559005b3bb318da01f505740adb0e782aaf14e/packages/istanbul-lib-instrument/test/specs/input-source-map.yaml#L5

Just add an argument that's something like `inputSourceMapClass: true`.

```
name: defined input source map
code: |
  output = "test"
inputSourceMapNotPojo: { file: "test.js", mappings: "", names: [], sourceRoot: undefined, sources: [ "test.js" ], sourcesContent: [ 'output = "test"' ], version: 3 }
tests:
  - name: sets the input source map
    args: []
    out: "test"
    lines: { '1': 1 }
    statements: { '0': 1 }
    inputSourceMap: { file: "test.js", mappings: "", names: [], sourceRoot: undefined, sources: [ "test.js" ], sourcesContent: [ 'output = "test"' ], version: 3 }
``` 

And in this file, https://github.com/istanbuljs/istanbuljs/blob/c1559005b3bb318da01f505740adb0e782aaf14e/packages/istanbul-lib-instrument/test/specs.test.js#L70

Create a:

```js
class NotPojo {
  constructor(obj) {
    Object.assign(this, obj)
  }
}
```

And create an instance of this if `inputSourceMapNotPojo` is set rather than `inputSourceMap`.

bcoe (changes requested):
Left some notes.

packages/istanbul-lib-instrument/src/visitor.js
marekdedic: Thanks for the suggestions!
- The simplification of the spread operator is a good idea, commited
- In the change suggestion you removed the part where the original sourceMap is restored - I took that over from #576, I am not sure if it's the right approach... Was that on purpose?
- I agree a test would be nice and I tried adding it, but I have no idea how the tests work (I've never used mocha, only jest...). I still have no idea.

packages/istanbul-lib-instrument/src/visitor.js
JLHwung: We can also print the constructor name in the warning message, so it is easier to identify what creates such input source map.

marekdedic: A good idea, thanks!

bcoe:
> In the change suggestion you removed the part where the original sourceMap is restored

I might be misreading, but I don't think this is necessary because you're updating the reference on `coverageData` already:
```js
> const foo = {inputSourceMap: {a: 99}}
undefined
> const otherValue = {b: 200}
undefined
> foo.inputSourceMap = {...otherValue}
{ b: 200 }
> foo
{ inputSourceMap: { b: 200 } }
```

> I agree a test would be nice and I tried adding it before submitting

I can add a test.

marekdedic:
Ok, thanks, I removed the restoring part. Could you please add the test? Thanks!

marekdedic:
@bcoe Could you please look at the test? It fails when I comment out the additions in this PR, so at least that checks out...

JamesCatt:
Just chiming in to ask what else is needed on this. Is it just approval from @bcoe ?

packages/istanbul-lib-instrument/test/specs.test.js
```diff
@@ -68,6 +74,11 @@ function generateTests(docs) {
                     const fn = async function() {
                         const genOnly = (doc.opts || {}).generateOnly;
                         const noCoverage = (doc.opts || {}).noCoverage;
+                        if (doc.inputSourceMapClass) {
```
bcoe, line 77: Thanks for the test 🥳

packages/istanbul-lib-instrument/src/visitor.js
```diff
@@ -785,6 +785,20 @@ function programVisitor(types, sourceFilePath = 'unknown.js', opts = {}) {
...
+            if (
+                coverageData.inputSourceMap &&
+                Object.getPrototypeOf(coverageData.inputSourceMap) !==
+                    Object.prototype
+            ) {
+                console.warn(
+                    'WARNING: istanbul-lib-instrument expects the source map to be a plain object. Instead, an object of class "' +
+                        coverageData.inputSourceMap.constructor.name +
+                        '" was passed.'
+                );
```
bcoe, line 797:
```suggestion
```

bcoe: I think this is a case where it would be fine to just silently recover.

bcoe (approved):

bcoe:
@marekdedic thank you for the contribution 👏

Issue #649, opened 2021-11-07
"E" is not showing in the HTML reporter for "implicit else" branches after https://github.com/istanbuljs/istanbuljs/pull/633


gareth-smyth:
**Expected Behavior: how should things be working?**
When an "implicit else" statement is not covered the HTML reporter should show an "E" to the left of the "if" statement.

![expected](https://user-images.githubusercontent.com/6697040/140655164-7c71f07d-1e1b-4fdf-9618-b2838ad0327a.png)

**Observed Behavior: what did you observe happening instead?**
The heading correctly shows that a branch is not covered but the branch is not highlighted in the code view.

![actual](https://user-images.githubusercontent.com/6697040/140655169-fe5701a8-9c51-4e61-8f48-52a2564c91db.png)

**Code: provide a repository, or snippet of code, that reproduces the bug.**
```
if(1 === 1) {
    console.log(1);
}
```

** Configuration **
Node v14.17.2
NPM 6.14.13
Running Jest 27.3.1

** Quick analysis **
If I revert the change from this PR https://github.com/istanbuljs/istanbuljs/pull/633 it seems to work correctly.

Thanks,
Gareth

bcoe:
@gareth-smyth thank you for the bug report  :+1: 

@adrian-burlacu-software any thoughts on this one?

Powerplex:
I can confirm we have the same issue. The generated json is correct, showing "0" lines covered for missing `else` branches. The static website generated to show the coverage is showing a coverage lower than 100% BUT does not show where it is. 

The [E] (Else branch not covered) is not showing for implicit else only. If we add a simple `else { /* with nothing inside */ }` then we can see the coverage is missing. 

@gareth-smyth this is the exact same issue.

adrian-burlacu-software:
@bcoe 
Sorry, this just hit my radar. I'll take a look to fix, hopefully by the end of the holiday break. Was lucky that I was Looking to make further improvements to this lib so I already made the time :)

Kind regards,
Adrian

adrian-burlacu-software:
@bcoe 
Still waiting to validate, my gut feeling is that it's a downstream issue. We fixed the reporting data but not the HTML reporter.

bcoe:
@adrian-burlacu-software :+1: thank you for starting to dig into this. I felt the recent additions to the library have added more value than this bug detracts, but would be nice to fix :smile:

adrian-burlacu-software:
@bcoe 
[...]

PR #663, opened 2022-01-05
Fix "E" is not showing in the HTML reporter for "implicit else" branches after pull 633

adrian-burlacu-software:
Fixes this issue:
[Issue 649](https://github.com/istanbuljs/istanbuljs/issues/649) after [Pull 633](https://github.com/istanbuljs/istanbuljs/pull/633)


packages/istanbul-reports/lib/html/annotator.js
```diff
@@ -164,6 +164,18 @@ function annotateBranches(fileCoverage, structuredText) {
                     gt;
                 closeSpan = lt + '/span' + gt;
 
+                if (
```
bcoe, line 167: I think this deserves a through comment, otherwise this looks good to me 😄

packages/istanbul-reports/test/html/annotator.js
```diff
@@ -81,5 +81,25 @@ describe('annotator', () => {
                 '<span class="fstat-no" title="function not covered" >  function test () {};</span>'
             );
         });
+
+        // see: https://github.com/istanbuljs/istanbuljs/issues/649
+        it('handles implicit else branches', () => {
```
bcoe, line 86: Thanks for the test 🥳

bcoe (changes requested):
Thanks for digging into this.

packages/istanbul-reports/lib/html/annotator.js
adrian-burlacu-software: Ok, I added a comment.

bcoe (approved):

PR #665, opened 2022-01-18
refactor: slightly improve PNG files compression

XhmikosR:
Before: 958 bytes
After:  721 bytes

There shouldn't be any visual difference, I compared the images with 500% zoom and they look identical.

Used zopflipng and optipng.


bcoe (approved):

bcoe (approved):

bcoe:
@XhmikosR thank you for the contribution.

Issue #670, opened 2022-02-08
Incorrect branch coverage when using implicit else post #633


alan-agius4:
Post #633 the implicit else is not being counted as a branch.

**Before**
![Screenshot 2022-02-08 at 10 10 18](https://user-images.githubusercontent.com/17563226/152954589-dd2f4235-04b0-48d6-8552-5e69149d0252.png)
```json
{
  "branchMap": {
    "0": {
      "loc": {
        "start": { "line": 2, "column": 2 },
        "end": { "line": 5, "column": null }
      },
      "type": "if",
      "locations": [
        {
          "start": { "line": 2, "column": 2 },
          "end": { "line": 5, "column": null }
        },
        {
          "start": { "line": 2, "column": 2 },
          "end": { "line": 5, "column": null }
        }
      ]
    }
  }
}
```
**After**
![Screenshot 2022-02-08 at 10 11 07](https://user-images.githubusercontent.com/17563226/152954717-1f180d57-b982-42dc-b91d-27eab09452cf.png)

```json
{
  "branchMap": {
    "0": {
      "loc": {
        "start": { "line": 2, "column": 2 },
        "end": { "line": 5, "column": null }
      },
      "type": "if",
      "locations": [
        {
          "start": { "line": 2, "column": 2 },
          "end": { "line": 5, "column": null }
        }
      ]
    }
  }
}
```
**Reproduction**

https://github.com/alan-agius4/coverage-issue-istanbul

```
git clone https://github.com/alan-agius4/coverage-issue-istanbul.git
cd coverage-issue-istanbul
yarn
yarn test
```

Reported in Angular CLI: https://github.com/angular/angular-cli/issues/22652

bcoe:
This was hopefully fixed by:

https://github.com/istanbuljs/istanbuljs/pull/663

Are you able to test against the latest release of the instrumenter?

CC: @adrian-burlacu-software

alan-agius4:
@bcoe, I am using the latest version and the issue is still persists.

luismartinez85:
@bcoe, I am using the latest version and the issue is still persists.

phyr0s:
Using lastest version and not working

adrian-burlacu-software:
@bcoe @alan-agius4 @phyr0s 

Ok so I originally fixed the coverage map in #633 because of the erroneous output.
Then, being a new contributor I figured downstream reporting would be updating by the rest of the genius crowd like yous guys :)
[...]

PR #679, opened 2022-04-01
fix: add placeholder to fix Implicit Else

adrian-burlacu-software:
@bcoe 
While implicit else reporting was fixed in pull #663, it no longer aligns to the instrumented API. Where I previously updated that to provide a placeholder in pull #633, by now the placeholder is gone. 

This became an issue #670 

The simplest was to update the reporting to reflect this change in the instrumenter API. Namely, I added a placeholder into the metadata array when the single if branch with no else condition.

I'd fix the design but that's already taking hours to reason about the various fails I could cause.


packages/istanbul-reports/test/fixtures/github-649.json
```diff
@@ -100,11 +100,6 @@
             "line": 6,
             "column": 5
           }
-        }, {
```
bcoe, line 103: I'm not understanding why you deleted this from your fixture?

Should we also handle the empty `start`/`end` use case?

packages/istanbul-reports/lib/html/annotator.js
```diff
@@ -142,6 +142,19 @@ function annotateBranches(fileCoverage, structuredText) {
         // only highlight if partial branches are missing or if there is a
         // single uncovered branch.
         if (sumCount > 0 || (sumCount === 0 && branchArray.length === 1)) {
+            // Need to recover the metaArray placeholder item to count an implicit else
+            if (
+                branchMeta[branchName].type === 'if' &&
+                branchArray.length === 2 &&
```
bcoe, line 148: I think adding a comment for what each of these lines reqpresents would be useful, does `branchArray.length === 2` imply that there's no `else?

bcoe (commented):
I left a couple questions, thanks for the patch.

packages/istanbul-reports/lib/html/annotator.js
adrian-burlacu-software: @bcoe
Will do Monday :)

Yes, the branchArray implies there's only the beginning if and the placeholder in the array.

bcoe:
@adrian-burlacu-software what's the state of this PR? I'm hesitant to land if it requires that we modify an existing regression test. But, since you wrote the regression test, you'd know better than me whether we should land.

adrian-burlacu-software:
> @adrian-burlacu-software what's the state of this PR? I'm hesitant to land if it requires that we modify an existing regression test. But, since you wrote the regression test, you'd know better than me whether we should land.

@bcoe 
Took a look at the files from pull #663  to resolve issue #649. It looks like I left out this placeholder in the test back then, meaning this should get landed.

bcoe (approved):

PR #685, opened 2022-06-01
Replace source-map with @jridgewell/trace-mapping

onigoetz:
The trace-mapping module is faster than source-map in every benchmark
https://github.com/jridgewell/trace-mapping#benchmarks

It's also being adopted by many packages across the ecosystem ([Jest](https://github.com/facebook/jest/pull/12692), [Babel](https://github.com/babel/babel/blob/main/CHANGELOG.md#v71710-2022-04-29), [Terser](https://github.com/terser/terser/blob/master/CHANGELOG.md#v5140) at least have included it already)
And it doesn't need to be async like `source-map` 0.7.*


packages/istanbul-lib-source-maps/test/map-store.test.js
```diff
@@ -90,7 +90,7 @@ describe('map store', () => {
                     },
                     end: {
                         line: 5,
-                        column: Infinity
+                        column: 1
```
onigoetz, line 93: I'm not the biggest expert in source-maps in general, my understanding is that in some edge cases `@jridgewell/trace-mapping` does have a different heuristic to resolve mappings.

Maybe @jridgewell can provide more details when this can happen

jridgewell: I think this comes from https://github.com/istanbuljs/istanbuljs/blob/e5d83d919dd1853f6382569082940f1d0bc14cc7/packages/istanbul-lib-source-maps/test/map-store.test.js#L137-L141?

If so, this appears to be the AST location for the `export` keyword, which ends at the ASI implied `;` after the closing `}` of the class declaration. It seems to record that location as `{ line: 5, column: 1 }`, since the `}` appears at column 0 of that line.

[Loading](https://evanw.github.io/source-map-visualization/#NzgwACJ1c2Ugc3RyaWN0IjsKdmFyIFNpbXBsZUNsYXNzID0gKGZ1bmN0aW9uICgpIHsKICAgIGZ1bmN0aW9uIFNpbXBsZUNsYXNzKCkgewogICAgfQogICAgU2ltcGxlQ2xhc3MucHJvdG90eXBlLmh5ID0gZnVuY3Rpb24gKCkgewogICAgICAgIGNvbnNvbGUubG9nKCJIeSIpOwogICAgfTsKICAgIHJldHVybiBTaW1wbGVDbGFzczsKfSgpKTsKZXhwb3J0cy5TaW1wbGVDbGFzcyA9IFNpbXBsZUNsYXNzOwovLyMgc291cmNlTWFwcGluZ1VSTD1kYXRhOmFwcGxpY2F0aW9uL2pzb247YmFzZTY0LGV3b2dJQ0oyWlhKemFXOXVJam9nTXl3S0lDQWlabWxzWlNJNklDSjBaWE4wTG1weklpd0tJQ0FpYzI5MWNtTmxVbTl2ZENJNklDSWlMQW9nSUNKemIzVnlZMlZ6SWpvZ1d3b2dJQ0FnSW5SbGMzUXVkSE1pQ2lBZ1hTd0tJQ0FpYzI5MWNtTmxjME52Ym5SbGJuUWlPaUJiQ2lBZ0lDQWlaWGh3YjNKMElHTnNZWE56SUZOcGJYQnNaVU5zWVhOeklIdGNiaUFnSUNCb2VTZ3BJSHRjYmlBZ0lDQWdJQ0FnWTI5dWMyOXNaUzVzYjJjb1hDSkllVndpS1R0Y2JpQWdJQ0I5WEc1OUlnb2dJRjBzQ2lBZ0ltNWhiV1Z6SWpvZ1cxMHNDaUFnSW0xaGNIQnBibWR6SWpvZ0lqdEJRVUZCTzBsQlFVRTdTVUZKUVN4RFFVRkRPMGxCU0Vjc2QwSkJRVVVzUjBGQlJqdFJRVU5KTEU5QlFVOHNRMEZCUXl4SFFVRkhMRU5CUVVNc1NVRkJTU3hEUVVGRExFTkJRVU03U1VGRGRFSXNRMEZCUXp0SlFVTk1MR3RDUVVGRE8wRkJRVVFzUTBGQlF5eEJRVXBFTEVsQlNVTTdRVUZLV1N4dFFrRkJWeXhqUVVsMlFpeERRVUZCSWdwOTM3OAB7CiAgInZlcnNpb24iOiAzLAogICJmaWxlIjogInRlc3QuanMiLAogICJzb3VyY2VSb290IjogIiIsCiAgInNvdXJjZXMiOiBbCiAgICAidGVzdC50cyIKICBdLAogICJzb3VyY2VzQ29udGVudCI6IFsKICAgICJleHBvcnQgY2xhc3MgU2ltcGxlQ2xhc3Mge1xuICAgIGh5KCkge1xuICAgICAgICBjb25zb2xlLmxvZyhcIkh5XCIpO1xuICAgIH1cbn0iCiAgXSwKICAibmFtZXMiOiBbXSwKICAibWFwcGluZ3MiOiAiO0FBQUE7SUFBQTtJQUlBLENBQUM7SUFIRyx3QkFBRSxHQUFGO1FBQ0ksT0FBTyxDQUFDLEdBQUcsQ0FBQyxJQUFJLENBQUMsQ0FBQztJQUN0QixDQUFDO0lBQ0wsa0JBQUM7QUFBRCxDQUFDLEFBSkQsSUFJQztBQUpZLG1CQUFXLGNBSXZCLENBQUEiCn0=) this into a visualizer, the answer becomes a little complicated. That closing `}` maps to 3 locations in the generated code:

Screen Shot 2022-06-01 at 9 31 05 PM

And the implied `;` maps to 4 locations:

Screen Shot 2022-06-01 at 9 30 29 PM

What's likely happened is that the old sourcemap code does not have a guaranteed lowest bound return value for its binary search. If you have multiple generated locations that originated from a source location, _any_ of those generated positions can be returned by the lookup depending on which one is hit first by the binary search. So if you had 1 extra mapping on that line, you would have gotten a different result.

With `TraceMap`, we always return the the left-most or right-most result when using `GREATEST_LOWER_BOUND` or `LEAST_UPPER_BOUND` (respectively). In this case, it seems we're returning a mapping that is on the same source line as the original mapping (in the `originalEndPositionFor` code). So we're detecting the real end bound correctly.

If you wanted more robust code, we could add a [allGeneratedPositionsFor](https://github.com/mozilla/source-map/#sourcemapconsumerprototypeallgeneratedpositionsfororiginalposition) API to return all the generated mapping locations. You could then iterate the array to find an appropriate one, and guarantee that you're always finding the real end location.

SimenB (approved):
thanks!

SimenB:
https://github.com/istanbuljs/istanbuljs/releases/tag/istanbul-lib-source-maps-v5.0.0

PR #687, opened 2022-06-19
Convert file url to path

NikolayMakhonin:
fix this bug: https://github.com/istanbuljs/nyc/issues/1473
It looks like typescript generated a source map with file urls instead the file paths.

The url.fileURLToPath function will work in Node >= v10.12.0
https://nodejs.org/api/url.html#urlfileurltopathurl


bcoe:
@NikolayMakhonin thank you for digging into this 👏 any chance we can add a test?

PR #688, opened 2022-06-28
fix: `new Date()` such that it works with MockDate library

uulwake:
Getting this error when running test coverage
```
TypeError: Class constructor MockDate cannot be invoked without 'new'
 ❯ new HtmlReport ../../node_modules/.pnpm/istanbul-reports@3.1.4/node_modules/istanbul-reports/lib/html/index.js:265:21
 ❯ Object.create ../../node_modules/.pnpm/istanbul-reports@3.1.4/node_modules/istanbul-reports/index.js:22:16
```


packages/istanbul-reports/lib/html/index.js
```diff
@@ -262,7 +262,7 @@ class HtmlReport extends ReportBase {
         this.verbose = opts.verbose;
         this.linkMapper = opts.linkMapper || standardLinkMapper;
         this.subdir = opts.subdir || '';
-        this.date = Date();
+        this.date = new Date();
```
bcoe, line 265: These are not equivalent types, `new Date()` returns a Date object, I believe `Date()` returns an ISO string of date.

Do you want `(new Date()).tostring()`?

bcoe (changes requested):

packages/istanbul-reports/lib/html/index.js
uulwake: Sure

bcoe (approved):

bcoe:
@uulwake thank you for the contribution.

PR #692, opened 2022-08-27
fix: support ESM when reading coverage

### LINKED ISSUE #614: Hardcoded Babel config prevents code coverage for ESM TypeScript modules
If I create an "src/index.ts" file with the following contents:
```
import { dirname } from 'path'
import { fileURLToPath } from 'url'

const __dirname = dirname(fileURLToPath(import.meta.url))

export function someFunction() {
   console.log(__dirname)
}
```

and run `jest --coverage`, I get:
```
STACK: SyntaxError: unknown: import.meta may appear only with 'sourceType: "module"' (3828:53)
Consider renaming the file to '.mjs', or setting sourceType:module or sourceType:unambiguous in your Babel config for this file.
```

Unfortunately I can't rename my file to "mjs" as it would not be a TypeScript file and would fail compilation.

This seems to be caused by the `packages/istanbul-lib-instrument/src/read-coverage.js` lines:
```
function getAst(code) {
    if (typeof code === 'object' && typeof code.type === 'string') {
        // Assume code is already a babel ast.
        return code;
    }

    if (typeof code !== 'string') {
        throw new Error('Code must be a string');
    }

    // Parse as leniently as possible
    return parseSync(code, {
        babelrc: false,
        configFile: false,
        parserOpts: {
            allowImportExportEverywhere: true,
            allowReturnOutsideFunction: true,
            allowSuperOutsideMethod: true,
            sourceType: 'script',
            plugins: defaults.instrumenter.parserPlugins
        }
    });
}
```

Because the "sourceType" is hard coded as 'script', there is no way to change it to use 'module' and no way to modify the babel config otherwise.

NotWearingPants:
Fixes #614

`read-coverage.js` attempts to parse the code into AST using Babel, and wants to `Parse as leniently as possible` according to the comment.
Changing the `sourceType` from `script` to `unambiguous` allows it to handle ESM code as well as regular scripts.


SimenB (approved):
This seems very reasonable to me

SimenB:
https://github.com/istanbuljs/istanbuljs/releases/tag/istanbul-lib-instrument-v6.0.1

PR #694, opened 2022-09-19
Add test case for json reporter

DevDengChao:


packages/istanbul-reports/test/json/index.js
```diff
@@ -0,0 +1,61 @@
...
+        file => {
+            if (file.indexOf('.json') !== -1) {
+                createTest(file);
+            }
+        }
+    );
+
+    // Some kind of strange race condition appears makes the test fail when running all test at once.
+    // But if you run this test case below alone, it passes.
+    // createTest('100-line-missing-branch.json');
```
DevDengChao, line 60: NEED HELP.
There are some kind of strange race condition appears makes the test fail when running all test at once.
But if you run the `100-line-missing-branch.json` test case alone, it passes.

PR #699, opened 2022-10-23
fix: wrong output `truecount` and `falsecount` of clover

### LINKED ISSUE #695: Branch counts of reporter may have a mistake
I'm using [codecov](https://about.codecov.io/) for code coverage. Codecov marked some lines as `Partial` means one of branches are not covered, but they are marked as covered in local HTML reports.

I think codecov are using `clover.xml` as the coverage information source. Then I checked my `clover.xml`, one of `truecount` and `falsecount` seems to be `0` when all branch are marked as covered in HTML reports.

---

I created [a repo](https://github.com/KiritaniAyaka/istanbul-coverage-report) for repro this. My lib are using vitest (vitest depends istanbul indirectly) so I created 2 branches for repro of istanbul and vitest.

**clover.xml:**

```xml
...
<line num="1" count="1" type="stmt"/>
<line num="3" count="2" type="cond" truecount="2" falsecount="0"/>
<line num="5" count="1" type="stmt"/>
<line num="9" count="1" type="stmt"/>
<line num="12" count="2" type="stmt"/>
...
```

**HTML:**

image

**Line 5, 9 are marked as covered in `clover.xml` , but the `falsecount` of branch at line 3 is `0`.**

**codecov:**

image

**Codecov think one of branches are not covered.**

I'm new for code coverage, so I'm not sure whether it is a bug of istanbul, or other lib, or codecov?

KiritaniAyaka:
close #695 

# Changes

- fix wrong `truecount` and `falsecount` in clover
- `switch` will be considered a statement now because it has no binary result (not applicable to `truecount` and `falsecount`).

### DIFF
--- a/packages/istanbul-reports/lib/clover/index.js
+++ b/packages/istanbul-reports/lib/clover/index.js
@@ -119,7 +119,7 @@ class CloverReport extends ReportBase {
     onDetail(node) {
         const fileCoverage = node.getFileCoverage();
         const metrics = node.getCoverageSummary();
-        const branchByLine = fileCoverage.getBranchCoverageByLine();
+        const branchDetails = getBranchDetails(fileCoverage);
 
         this.xml.openTag('file', {
             name: asClassName(node),
@@ -135,12 +135,12 @@ class CloverReport extends ReportBase {
                 count,
                 type: 'stmt'
             };
-            const branchDetail = branchByLine[k];
+            const branchDetail = branchDetails[k];
 
-            if (branchDetail) {
+            if (branchDetail && branchDetail.type === 'if') {
                 attrs.type = 'cond';
-                attrs.truecount = branchDetail.covered;
-                attrs.falsecount = branchDetail.total - branchDetail.covered;
+                attrs.truecount = branchDetail.states[0];
+                attrs.falsecount = branchDetail.states[1];
             }
             this.xml.inlineTag('line', attrs);
         });
@@ -159,6 +159,22 @@ function asJavaPackage(node) {
 
 function asClassName(node) {
     return node.getRelativeName().replace(/.*[\\/]/, '');
+}
+
+function getBranchDetails(fileCoverage) {
+    const branchMeta = fileCoverage.branchMap;
+    const branchStats = fileCoverage.b;
+
+    const branchDetails = {};
+
+    Object.entries(branchMeta).forEach(([index, branch]) => {
+        branchDetails[branch.loc.start.line] = {
+            type: branch.type,
+            states: branchStats[index]
+        };
+    });
+
+    return branchDetails;
 }
 
 module.exports = CloverReport;
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ dist
 lerna-debug.log
 packages/*/yarn.lock
 packages/nyc/
+.vscode/*


packages/istanbul-reports/lib/clover/index.js
```diff
@@ -135,12 +135,12 @@ class CloverReport extends ReportBase {
                 count,
                 type: 'stmt'
             };
-            const branchDetail = branchByLine[k];
+            const branchDetail = branchDetails[k];
 
-            if (branchDetail) {
+            if (branchDetail && branchDetail.type === 'if') {
```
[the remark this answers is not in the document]
KiritaniAyaka: Oh, thanks for point out. I'll dig it further later.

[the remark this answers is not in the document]
KiritaniAyaka: I have done a simple search in source code and a testing (Check it in [my repo](https://github.com/KiritaniAyaka/istanbul-coverage-report/blob/main/test/index.test.ts#L27-L49)). Except `if ` and `switch`, there are three types of branches:

`default-arg`: 
The `states` always have one member and it presents the number of times that default arguments are applied.

`cond-expr`: (ternary operator)
The `states` always have two members. The first presents the number of times that condition was judged as `true`, and the second presents the number of times that condition was judged as `false`.

`binary-expr`:
The number of members is equal to the number of conditions. The xth member presents the number of times that xth condition was calculated. And it has nothing to do with the nest and calculating order.

---

For `default-arg`, I think it should be considered a `statement` according to [the standard](https://gist.github.com/aik099/8556655#file-clover-xsd-L117-L119) mentioned at #695 .

For `cond-expr`, it could be deal with the same as `if`.

I have no good idea to deal with `binary-expr`. Maybe we can consider it as covered when all conditions are calculated?

KiritaniAyaka:
I dealt with the remaining branch types in the latest commit. Please check again. @bcoe

silverbackdan:
Would be a great and very useful fix, I'm in no position to reproduce but have been experiencing this issue with partial matches. Thanks all for the work leading to this PR. In the linked issue I have another simple reproduction if required.

packages/istanbul-reports/lib/clover/index.js
```diff
@@ -135,13 +135,35 @@ class CloverReport extends ReportBase {
...
+                attrs.falsecount = branchDetail.states[1];
+            } else if (
+                ['binary-expr', 'default-arg'].includes(branchDetail.type)
+            ) {
+                if (branchDetail.states.every(state => state > 0)) {
+                    attrs.truecount = 1;
+                    attrs.falsecount = 1;
+                } else {
+                    attrs.truecount = 1;
+                    attrs.falsecount = 0;
```
KiritaniAyaka, line 163: Hi, @bcoe. The problem mentioned in issue #695 looks to make others puzzled now so I think it's needed to push this PR.

I guess this PR stop forward because you may disagree with the practice that assigning hardcode values to make 3rd-party parsers understand if this line was covered or not.

Could we have better practice to resolve this problem?

silverbackdan:
FYI if anyone finds this thread and is looking for an answer, I've found uploading coverage-final.json instead of clover.xml to codecov results in the correct results and there are not partial line matches galore.

KiritaniAyaka:
> FYI if anyone finds this thread and is looking for an answer, I've found uploading coverage-final.json instead of clover.xml to codecov results in the correct results and there are not partial line matches galore.

Great alternative plan!

Just enable json reporter in `.nycrc` to resolve it, for example:

```json
{
  "extends": "@istanbuljs/nyc-config-typescript",
  "include": ["src/**/*.ts"],
  "extension": [".ts"],
  "reporter": ["html", "json"], // enable json reporter
  "all": true,
  "sourceMap": true,
  "instrument": true
}
```

Here's my [coverage report on codecov](https://app.codecov.io/github/KiritaniAyaka/istanbul-coverage-report/commit/8e87fd2686faf9be0b359fddfae23509068a5490/blob/src/index.ts) after switch to json reporter

silverbackdan:
Great, pleased it's helped someone. I found the idea from uploading a different type of report for a different coverage provider so thought it worth a shot. I'd like to my updated coverage to show as well, but it's on a feature branch which will be deleted shortly, but anyway. Happy days.

packages/istanbul-reports/lib/clover/index.js
```diff
@@ -135,13 +135,35 @@ class CloverReport extends ReportBase {
...
                 type: 'stmt'
             };
-            const branchDetail = branchByLine[k];
+            const branchDetail = branchDetails[k];
 
-            if (branchDetail) {
-                attrs.type = 'cond';
-                attrs.truecount = branchDetail.covered;
-                attrs.falsecount = branchDetail.total - branchDetail.covered;
+            if (!branchDetail || branchDetail.type === 'switch') {
```
bcoe, line 140: What is the effect of adding a `line` tag here and returning immediately, will this result in 100% line coverage and no missing branch coverage?

packages/istanbul-reports/lib/clover/index.js
```diff
@@ -135,13 +135,35 @@ class CloverReport extends ReportBase {
...
+
+            attrs.type = 'cond';
+            attrs.truecount = 0;
+            attrs.falsecount = 0;
+
+            if (count === 0) {
+                return this.xml.inlineTag('line', attrs);
+            }
+
+            if (['if', 'cond-expr'].includes(branchDetail.type)) {
```
bcoe, line 152: This logic is complex enough, that I think it would be good to add a description above each condition, explaining the reasoning for the logic.

bcoe (changes requested):
Your implementation seems good to me, but it would be helpful to have a comment above each logical check, to explain the decisions you've made.

packages/istanbul-reports/lib/clover/index.js
KiritaniAyaka: It is from the [original code](https://github.com/istanbuljs/istanbuljs/blob/fb8cb4bb99c9e5d36c8b011338d2ab5d3c72ff9c/packages/istanbul-reports/lib/clover/index.js#L140-L145):

```javascript
if (branchDetail) {
    attrs.type = 'cond';
    attrs.truecount = branchDetail.covered;
    attrs.falsecount = branchDetail.total - branchDetail.covered;
}
this.xml.inlineTag('line', attrs);
```

but the true case is more complex so I move the false case into the brackets and change the condition to `!branchDetail`.

As to `branchDetail.type === 'switch'`. Thanks to you, I found a situation as you said: 

```javascript
switch (num) {
    case 0:
    case 1:
        console.log('Hello');
        break;
}
```

When we use num=1 to test this switch, all lines are covered but branch not.

Maybe we should treat `switch` the same as `binary-expr`.

PR #704, opened 2022-11-03
chore: use npm workspaces

### LINKED ISSUE #703: monorepo improvement: use npm/pnpm workspaces
Instead of [manual script](https://github.com/istanbuljs/istanbuljs/blob/master/package.json#L10) 
would be great to use native monorepo utils to install npm dependencies.

We can use just [npm workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) or [pnpm workspaces](https://pnpm.io/workspaces) to manage dependencies.

This will help to have a easier way to link internal packages between them, same for bump versions.

sebaplaza:
Closes #703

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ _You're probably actually looking for one of the following repos:_
 
 Contributing to the packages contained in this repo is easy:
 
-1. after checking out, run `npm install` (this will run the lerna build).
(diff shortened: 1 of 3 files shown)


SimenB:
This needs to upgrade npm on Ci seemingly

PR #720, opened 2023-04-29
cobertura: emit line coverage for each method

RoadTrain:
Cobertura is usually structured in way that line coverage is presented for both classes and methods. See https://github.com/danielpalme/ReportGenerator/issues/603 for more context.

This PR improves Cobertura generation so that for each method there is separate line coverage data.

### DIFF
--- a/packages/istanbul-reports/lib/cobertura/index.js
+++ b/packages/istanbul-reports/lib/cobertura/index.js
@@ -79,7 +79,42 @@ class CoberturaReport extends ReportBase {
     onDetail(node) {
         const fileCoverage = node.getFileCoverage();
         const metrics = node.getCoverageSummary();
+        const lines = fileCoverage.getLineCoverage();
         const branchByLine = fileCoverage.getBranchCoverageByLine();
+
+        const emitLines = (startLine, endLine) => {
+            this.xml.openTag('lines');
+
+            Object.entries(lines).forEach(([k, hits]) => {
+                if (
+                    (startLine !== undefined && k < startLine) ||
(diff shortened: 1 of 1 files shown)


RoadTrain:
TODO: Fix tests

bcoe:
@RoadTrain I have never used the Cobertura format myself and am open to suggestions if you can get tests passing. I would also like to make sure that we don't break tests in major dependencies like c8 and nyc.

RoadTrain:
A quick update. I will try to get back to it soon. Some tests were broken even before my changes, so I have to makes sure they are fixed in a correct way.

RoadTrain:
For future reference, there's an official Cobertura spec, which specifies line data for methods: https://github.com/cobertura/cobertura/blob/master/cobertura/src/site/htdocs/xml/coverage-04.dtd

The original `cobertura` tool (for java projects) emits methods with all line data corresponding to each particular method: https://github.com/cobertura/cobertura/blob/0ff963284cecaace30f409a977dccb07c41a5a8f/cobertura/src/main/java/net/sourceforge/cobertura/reporting/xml/XMLReport.java#L215

C# `coverlet` cobertura reporter does the same: https://github.com/coverlet-coverage/coverlet/blob/cceb71d758a57425ef54ca6bf9f0e443dfe378c5/src/coverlet.core/Reporters/CoberturaReporter.cs#L125

So this PR is basically bringing `instanbuljs` cobertura reporter in line with other reporters.

Hopefully I can get it finished after tests are fixed (https://github.com/istanbuljs/istanbuljs/pull/748).

PR #728, opened 2023-07-05
fix(deps): update semver to address CVE-2022-25883

super132:
To fix the [security vulnerability](https://github.com/advisories/GHSA-c2qf-rxjj-qqgw) of `semver` `v6.x`. Related to #727.

PR #731, opened 2023-07-25
fix(deps)!: upgrade semver CVE-2022-25883

bcoe:
BREAKING CHANGE: minimum Node version bumped to 10.


bcoe:
#728 needs to be broken into multiple updates due to engine version.

PR #741, opened 2023-09-08
bump node version to 16

### LINKED ISSUE #739: Upgrade Node.js LTS Version and package.json Update
I suggest that we consider upgrading the existing Node.js LTS version within the Istanbul project. This would allow us to take advantage of the latest enhancements and security updates. Additionally, it's worth considering an update to our package.json file to specify a higher `engine.node` version.

**Advantages:**
1. **Improved Performance:** The newer LTS versions come with optimizations that can significantly enhance the speed of our code execution.

2. **Enhanced Security:** An upgrade would ensure that we promptly address any potential vulnerabilities, thereby upholding the trust of our users.

3. **Access to Modern Features:** This upgrade would grant us access to new features of Node.js, facilitating more efficient development and the creation of cleaner code.

4. **Sustained Support:** LTS versions provide a stable foundation with ongoing support, minimizing the need for disruptive upgrades.

5. **Streamlined Collaboration:** Using a common Node.js version simplifies collaboration and aids in the smooth onboarding of contributors.

6. **Removal of dependencies/polyfills: ** with new access to features of Node.js we don't need as many dependencies

jimmywarting:
This update raises the minimum required Node.js version to v16. This change enables us to embrace newer technologies without hesitation and eliminates the need to consider whether it's acceptable to utilize modern solutions or resort to older code styles. This enhancement not only improves the overall development experience but also encourages contributions if they can contribute with newer modern code.

fix #739


SimenB:
This should be fine to do, but I'm not sure if a breaking change _just_ to bump the node version is needed (unless it also allows us to e.g. update some dependencies or some such)

bcoe:
@jimmywarting bother you to update this to Node 18, now that 16 is EOL?

@SimenB given we're jumping so many version #s as a minimum engine #, I think this should be a breaking change (upstream users can choose not to bump their major).

SimenB:
yeah, definitely a breaking change regardless of which versions are dropped. My question was more - what does it give us to drop old versions?

PR #743, opened 2023-09-09
fix: replace `source-map` with `@jridgewell/trace-mapping`

SimenB:
Following up https://github.com/istanbuljs/v8-to-istanbul/pull/186

I don't think this is visible to consumers?

/cc @jridgewell FYI 🙂


packages/istanbul-lib-source-maps/lib/map-store.js
```diff
@@ -190,15 +190,17 @@ class MapStore {
...
+                    const smc = new TraceMap(obj);
                     smc.sources.forEach(s => {
-                        const content = smc.sourceContentFor(s);
-                        if (content) {
-                            const sourceFilePath = pathutils.relativeTo(
-                                s,
-                                filePath
-                            );
-                            this.sourceStore.set(sourceFilePath, content);
+                        if (s) {
```
SimenB, line 195: [types](https://github.com/jridgewell/trace-mapping/blob/5ccfcfeeee9dfa3b13567bb0f95260ea32f2c269/src/types.ts#L8) says `s` is `string | null`

packages/istanbul-lib-source-maps/test/map-store.test.js
```diff
@@ -90,7 +90,7 @@ describe('map store', () => {
                     },
                     end: {
                         line: 5,
-                        column: Infinity
+                        column: 1
```
SimenB, line 93: not sure if good or bad?

packages/istanbul-lib-source-maps/lib/map-store.js
jridgewell: Correct. If sources content is enabled, some tools will use `null` if the file's source content isn't available. So some files in the array will have source content (a string), and some won't (null).

packages/istanbul-lib-source-maps/test/map-store.test.js
jridgewell: What is the source file, output file, and source map? I don't return an `Infinity` value in `trace-mapping`, and I don't know what computes `end` here.

SimenB: I don't really know this code base, so the best answer I have for your q is "this test" 😅

AriPerkkio (approved):
I tested these changes on Vitest's internal coverage tests (V8 and Istanbul) by patching the package. All tests pass like usual and coverage reports are identical.

Anything missing from this PR? Happy to see `source-map` replaced with lighter package.

SimenB:
Missing a review pretty much - I cannot self review 😅

packages/istanbul-lib-source-maps/test/map-store.test.js
SimenB: I've gotten admin access to this repo, so I'm able to merge this now. I'd like to verify this one, tho. So if you're able to take a look, I'd be grateful!

>What is the source file, output file, and source map?

I believe there are present in a comment right below:

https://github.com/istanbuljs/istanbuljs/blob/5584b50305a6a17d3573aea25c84e254d4a08b65/packages/istanbul-lib-source-maps/test/map-store.test.js#L136-L302

Input lines 136-141, output 142-152 and `inputSourceMap` is lines 293-301

SimenB: The origin of `Infinity` I'm guessing is from: https://github.com/istanbuljs/istanbuljs/blob/5584b50305a6a17d3573aea25c84e254d4a08b65/packages/istanbul-lib-source-maps/lib/get-mapping.js#L57-L72

bcoe (commented):
@SimenB on vacation and am starting to look through a few old PRs.

This update looks innocent enough to me, but I'm worried about the change from Infinity to 1 too ... and admittedly don't know this part of the codebase too well 😅 

I think this might be able to left open for a bit until someone has the time to dig in deeply.

packages/istanbul-lib-source-maps/test/map-store.test.js
SimenB: @jridgewell sorry to ping you again, but does the added context above help in evaluating if the change is correct (or at least not regressive) or not?

jridgewell: All good, I missed this during Thanksgiving week.

Based on this, the visualization we're tracing is [this](https://evanw.github.io/source-map-visualization/#NzIxACJ1c2Ugc3RyaWN0IjsKdmFyIFNpbXBsZUNsYXNzID0gKGZ1bmN0aW9uICgpIHsKICAgIGZ1bmN0aW9uIFNpbXBsZUNsYXNzKCkgewogICAgfQogICAgU2ltcGxlQ2xhc3MucHJvdG90eXBlLmh5ID0gZnVuY3Rpb24gKCkgewogICAgICAgIGNvbnNvbGUubG9nKCJIeSIpOwogICAgfTsKICAgIHJldHVybiBTaW1wbGVDbGFzczsKfSgpKTsKZXhwb3J0cy5TaW1wbGVDbGFzcyA9IFNpbXBsZUNsYXNzOwovLyMgc291cmNlTWFwcGluZ1VSTD1kYXRhOmFwcGxpY2F0aW9uL2pzb247YmFzZTY0LGV5SjJaWEp6YVc5dUlqb3pMQ0ptYVd4bElqb2lkR1Z6ZEM1cWN5SXNJbk52ZFhKalpWSnZiM1FpT2lJaUxDSnpiM1Z5WTJWeklqcGJJblJsYzNRdWRITWlYU3dpYzI5MWNtTmxjME52Ym5SbGJuUWlPbHNpWlhod2IzSjBJR05zWVhOeklGTnBiWEJzWlVOc1lYTnpJSHRjYmlBZ0lDQm9lU2dwSUh0Y2JpQWdJQ0FnSUNBZ1kyOXVjMjlzWlM1c2IyY29YQ0pJZVZ3aUtUdGNiaUFnSUNCOVhHNTlJbDBzSW01aGJXVnpJanBiWFN3aWJXRndjR2x1WjNNaU9pSTdRVUZCUVR0SlFVRkJPMGxCU1VFc1EwRkJRenRKUVVoSExIZENRVUZGTEVkQlFVWTdVVUZEU1N4UFFVRlBMRU5CUVVNc1IwRkJSeXhEUVVGRExFbEJRVWtzUTBGQlF5eERRVUZETzBsQlEzUkNMRU5CUVVNN1NVRkRUQ3hyUWtGQlF6dEJRVUZFTEVOQlFVTXNRVUZLUkN4SlFVbERPMEZCU2xrc2JVSkJRVmNzWTBGSmRrSXNRMEZCUVNKOQozMzMAeyJ2ZXJzaW9uIjozLCJmaWxlIjoidGVzdC5qcyIsInNvdXJjZVJvb3QiOiIiLCJzb3VyY2VzIjpbInRlc3QudHMiXSwic291cmNlc0NvbnRlbnQiOlsiZXhwb3J0IGNsYXNzIFNpbXBsZUNsYXNzIHtcbiAgICBoeSgpIHtcbiAgICAgICAgY29uc29sZS5sb2coXCJIeVwiKTtcbiAgICB9XG59Il0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiI7QUFBQTtJQUFBO0lBSUEsQ0FBQztJQUhHLHdCQUFFLEdBQUY7UUFDSSxPQUFPLENBQUMsR0FBRyxDQUFDLElBQUksQ0FBQyxDQUFDO0lBQ3RCLENBQUM7SUFDTCxrQkFBQztBQUFELENBQUMsQUFKRCxJQUlDO0FBSlksbUJBQVcsY0FJdkIsQ0FBQSJ9).

The column that's changed is the end location for the anonymous IIFE. The mappings on line 9 are (using 1-based line and 0-based column as `source-map` does):
1. `{ genColumn: 0, sourcesIndex: 0, sourceLine: 5, sourceColumn: 0 }` 
2. `{ genColumn: 1, sourcesIndex: 0, sourceLine: 5, sourceColumn: 1 }`
3. `{ genColumn: 1, sourcesIndex: 0, sourceLine: 1, sourceColumn: 0 }`
4. `{ genColumn: 5, sourcesIndex: 0, sourceLine: 5, sourceColumn: 1 }`

Notice that `genColumn` 1 has 2 mappings. In `trace-mapping` when doing the [originalPositionFor search](https://github.com/istanbuljs/istanbuljs/blob/72aa1fbda9e7aa49609a348a001b415cb962f4b2/packages/istanbul-lib-source-maps/lib/get-mapping.js#L84-L88) with `{ line: 9, column: 1 }`, the first specified is returned (due to `GREATEST_LOWER_BOUND`). So [beforeEndMapping](https://github.com/istanbuljs/istanbuljs/blob/72aa1fbda9e7aa49609a348a001b415cb962f4b2/packages/istanbul-lib-source-maps/lib/get-mapping.js#L37-L41) points to `{ line: 5, column: 1 }` in the original code. We then run a [generatedPositionFor](https://github.com/istanbuljs/istanbuljs/blob/72aa1fbda9e7aa49609a348a001b415cb962f4b2/packages/istanbul-lib-source-maps/lib/get-mapping.js#L51-L56) search, returning `{ line: 10, column: 34 }`. Note that the last seen generated position that points to that source location is used (because of the `LEAST_UPPER_BOUND` bias), and there are many in this example.

However `source-map` uses a first-match style binary search. So our `originalPostitionFor` returns **any** match for the generated position `{ line: 9, column: 1 }`. In this case, it hits `{ genColumn: 1, sourcesIndex: 0, sourceLine: 1, sourceColumn: 0 }` first. Doing the `generatedPositionFor` with that position gives `{ line: 2, column: 0 }`.

I dislike the results that `source-map` gives when there are multiple matches. If there were a different number of mappings on that line, then a different original position could be hit by the binary search.

I know we've discussed `originalEndPositionFor` in other PRs, but I can't find it now. The way that you're doing this kinda buggy. You're returning a single generated position out of the mapping, and that generated position could be sourced any line (hence the [line check](https://github.com/istanbuljs/istanbuljs/blob/72aa1fbda9e7aa49609a348a001b415cb962f4b2/packages/istanbul-lib-source-maps/lib/get-mapping.js#L64-L65)). In a case like this, you could incorrectly return an `Infinity` end when there _could_ be more code on that source line that would actually end that range. What you really want is to get the list of every generated position corresponding to that source location, and see if any of them are on the correct line. Ie, call `allGeneratedPositionsFor`

jridgewell: Oh hey, it was https://github.com/istanbuljs/istanbuljs/pull/685. Looks like my opinion hasn't changed since I wrote that, we should be using `allGeneratedPositionsFor`. 😄

SimenB: Oh, I completely missed #685! I'll land that then (after your new PR). Thank you!

SimenB:
Duplicate of #685

SimenB:
https://github.com/istanbuljs/istanbuljs/releases/tag/istanbul-lib-source-maps-v5.0.0

Issue #552, opened 2020-04-04
cobertura reporter does not construct package name attributes from projectRoot


tsondergaard:
The image below shows an excerpt of a coverage report displayed by the jenkins CI cobertura plugin for a project that is mostly implemented in C++, but which has a component written in TypeScript running in node that is tested with jest using the istanbul cobertura coverage reporter. In the picture "sparrow" is the name of a component written in C++. The "src", "src.controller", "src.db", etc packages are all from the node subproject. My issue is that these packages don't have the subproject name prefixed.

![image](https://user-images.githubusercontent.com/2016842/78461169-0e608f00-76c7-11ea-941d-661b5e95cc7c.png)

I pass a projectRoot config property to jest which istanbul-reports/lib/cobertura/index.js uses for filepaths, but not for package names. This is what is generated:

```xml
<?xml version="1.0" ?>
<!DOCTYPE coverage SYSTEM "http://cobertura.sourceforge.net/xml/coverage-04.dtd">
<coverage lines-valid="3046" lines-covered="1582" line-rate="0.5194" branches-valid="1355" branches-covered="531" branch-rate="0.39189999999999997" timestamp="1586032913784" complexity="0" version="0.1">
  <sources>
    <source>/home/foo/src/main-project</source>
  </sources>
  <packages>
    <package name="src" line-rate="0.7398" branch-rate="0.2386">
      <classes>
        <class name="app.ts" filename="sub-project/src/app.ts" line-rate="0.8934000000000001" branch-rate="0.16">
         ....
```

The problem is the package attribute `name="src"` it should be `name="sub-project.src"` to be consistent with the projectRoot and match the path in the `class filename="sub-project/src/app.ts"` attribute.

bitcloud:
I don't think that this issue can bis solved in the cobertura reporter as it depends on `getRelativeName` from `SummarizerFactory`.

`SummarizerFactory` try's to detect the project root by automatically detecting a common base directory.

```
/project/src/package1/subpackage1/index.js
/project/src/package1/subpackage2/index.js
----
commonParent => /project/src/package1
```

I guess `SummarizerFactory` needs to be projectRoot aware to fix this.

PR #748, opened 2023-10-10
fix: cobertura paths fix

Aeolun:
This PR does 3 things:

- Re-enables the test for all coverage reporters. I think at some point during a switch from lerna the whole thing was accidentally disabled.
- Modifies the Cobertura coverage reporter to fix issue #552 (basically, output package paths relative to the specified source directory).
- Upgrades the action versions in the `.github/workflows/ci.yml` file, so Github stops spewing errors when running them.


packages/istanbul-reports/test/lcovonly/index.js
```diff
@@ -26,23 +26,30 @@ describe('LcovOnlyReport', () => {
...
-            const context = istanbulLibReport.createContext({
-                dir: './',
-                coverageMap: istanbulLibCoverage.createCoverageMap(fixture.map)
-            });
-            const tree = context.getTree('pkg');
-            const report = new LcovOnlyReport(fixture.opts);
-            tree.visit(report, context);
-            const output = FileWriter.getOutput().replace(/SF:.*/, 'SF:');
-            if (fixture.lcovonlyExpected) {
+        if (fixture.lcovonlyExpected) {
```
bcoe, line 29: This seems a little strange, why are we only running these next lines for the one specific fixture?

PR #639, opened 2021-10-12
Report Logic Truthiness: Port over coverage lib changes.

adrian-burlacu-software:
@bcoe 
This are the istanbul-lib-coverage changes to do with evaluating logical expression truthiness.
Blocking the [original pull request ](https://github.com/istanbuljs/istanbuljs/pull/629) until the version is published.

All tests pass.

PR #750, opened 2023-11-02
fix: merge ranges properly when contained by other ranges in set

isaacs:
Fix: https://github.com/istanbuljs/v8-to-istanbul/issues/233

### DIFF
--- a/packages/istanbul-lib-coverage/lib/file-coverage.js
+++ b/packages/istanbul-lib-coverage/lib/file-coverage.js
@@ -45,35 +45,122 @@ function assertValidObject(obj) {
 const keyFromLoc = ({ start, end }) =>
     `${start.line}|${start.column}|${end.line}|${end.column}`;
 
+// When merging, we can have a case where two ranges cover
+// the same block of code with `hits=1`, and each carve out a
+// different range with `hits=0` to indicate it's uncovered.
+// Find the nearest container so that we can properly indicate
+// that both sections are hit.
+// Returns null if no containing item is found.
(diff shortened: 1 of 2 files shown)


SimenB:
@isaacs I'm not allowed to push to this branch - can you run `npm run fix` locally?

packages/istanbul-lib-coverage/lib/file-coverage.js
```diff
@@ -320,7 +411,7 @@ class FileCoverage {
         ret.branches = this.computeBranchTotals('b');
         // Tracking additional information about branch truthiness
         // can be optionally enabled:
-        if (this['bt']) {
```
SimenB, line 323: huh, typo before?

SimenB (approved):
this LGTM and only adds tests without changing old ones, so I think this is good to go 👍 

Need to fix prettier for CI, tho

packages/istanbul-lib-coverage/lib/file-coverage.js
isaacs: Seems that way!

isaacs: If this was intentional, then the code should have a test, because it wasn't being covered.

SimenB: digging through #639 and linked efforts, seems like this was simply a typo

isaacs:
Fixed lint errors, PTAL, thanks :)

SimenB (approved):

Issue #753, opened 2023-11-06
[Bug] Unable to merge coverage for the same file


sashuk:
One of my projects started failing with the following message (I use latest https://github.com/ljharb/istanbul-merge that pulls latest https://www.npmjs.com/package/istanbul-lib-coverage)

```bash
aleksandr@Alexanders-MacBook-Pro-2 test-portal % npx istanbul-merge --out "coverage/coverage-final-1.json" "./**/coverage/coverage-final.json"
/Users/aleksandr/.npm/_npx/98709a0041747bf7/node_modules/istanbul-lib-coverage/lib/file-coverage.js:74
            itemLoc.start.line - mapLoc.start.line,
                    ^

TypeError: Cannot read properties of undefined (reading 'start')
    at findNearestContainer (/Users/aleksandr/.npm/_npx/98709a0041747bf7/node_modules/istanbul-lib-coverage/lib/file-coverage.js:74:21)
    at addNearestContainerHits (/Users/aleksandr/.npm/_npx/98709a0041747bf7/node_modules/istanbul-lib-coverage/lib/file-coverage.js:121:23)
    at mergeProp (/Users/aleksandr/.npm/_npx/98709a0041747bf7/node_modules/istanbul-lib-coverage/lib/file-coverage.js:147:25)
    at FileCoverage.merge (/Users/aleksandr/.npm/_npx/98709a0041747bf7/node_modules/istanbul-lib-coverage/lib/file-coverage.js:301:27)
    at CoverageMap.addFileCoverage (/Users/aleksandr/.npm/_npx/98709a0041747bf7/node_modules/istanbul-lib-coverage/lib/coverage-map.js:112:29)
    at /Users/aleksandr/.npm/_npx/98709a0041747bf7/node_modules/istanbul-lib-coverage/lib/coverage-map.js:55:18
    at Array.forEach (<anonymous>)
    at CoverageMap.merge (/Users/aleksandr/.npm/_npx/98709a0041747bf7/node_modules/istanbul-lib-coverage/lib/coverage-map.js:54:35)
    at /Users/aleksandr/.npm/_npx/98709a0041747bf7/node_modules/istanbul-merge/bin/istanbul-merge:36:6
    at forEachArray (/Users/aleksandr/.npm/_npx/98709a0041747bf7/node_modules/for-each/index.js:12:17)

Node.js v18.16.0
```

It started failing right after https://github.com/istanbuljs/istanbuljs/pull/750/files was published as `latest`.

[...]

PR #754, opened 2023-11-06
fix: do not throw on missing range locations

isaacs:
Fix: #753

### DIFF
--- a/packages/istanbul-lib-coverage/lib/file-coverage.js
+++ b/packages/istanbul-lib-coverage/lib/file-coverage.js
@@ -45,6 +45,12 @@ function assertValidObject(obj) {
 const keyFromLoc = ({ start, end }) =>
     `${start.line}|${start.column}|${end.line}|${end.column}`;
 
+const isObj = o => !!o && typeof o === 'object';
+const isLineCol = o =>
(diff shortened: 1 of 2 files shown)


isaacs:
@sashuk It needs a review from a collaborator who didn't write it (ie, not from me), and I believe someone to publish the resulting version, if that isn't set up to work automatically.

I do have commit rights here, but I haven't really been terribly involved with this project, so I don't want to overstep unless @SimenB or @bcoe or someone asks me to.

SimenB:
I always feel like I'm overstepping when doing stuff in this project, but as all the original maintainers are busy, we do what we gotta do 😅

SimenB (approved):
Probably more defensive than it needs to be, but nothing wrong with that (especially considering the regression 😅)

isaacs:
@SimenB it's your kitten now, no returns no refunds! 😅

Re the defensiveness, yeah, I mean, this is a perfect example of the kind of error that TS would have clearly prevented. I've gotten so used to relying on types for this sort of thing, got kinda lazy in that last patch.  But without them, it's like everything is `any`, and I feel like it's a good idea to assume the worst.

sashuk:
Thank you all!

PR #706, opened 2022-11-04
fix: `istanbul-lib-source-maps` to preserve implicit `else` when sourcemaps are used

AriPerkkio:
- Fixes #705 

Tested the fix with reproduction case from downstream (https://github.com/vitest-dev/vitest/issues/2239) by modifying `./node_modules/istanbul-lib-source-maps/lib/transformer.js` on the fly. 

```
----------|---------|----------|---------|---------|-------------------
File      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
----------|---------|----------|---------|---------|-------------------
All files |     100 |       50 |     100 |     100 |                   
 cov.js   |     100 |       50 |     100 |     100 | 2                 
----------|---------|----------|---------|---------|-------------------

```

PR #789, opened 2024-07-01
fix: `istanbul-lib-source-maps` implicit `else` crash edge case

AriPerkkio:
- Fixes https://github.com/istanbuljs/istanbuljs/pull/706#issuecomment-2200012354

In Vitest codebase the `istanbul-lib-source-maps@5.0.5` caused a crash in edge case. The crash happend when `swc` transpiled code was added to report as uncovered file (`coverage.all`).

We can safely look for `source` before adding implicit `else` into the map. There should always be `source` already set from the previous loop of the `if` itself.

### DIFF
--- a/packages/istanbul-lib-source-maps/lib/transformer.js
+++ b/packages/istanbul-lib-source-maps/lib/transformer.js
@@ -84,6 +84,7 @@ class SourceMapTransformer {
                 }
                 // Check if this is an implicit else
                 else if (
+                    source &&
                     branchMeta.type === 'if' &&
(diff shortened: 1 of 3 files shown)


SimenB:
Do we need a test in this case?

AriPerkkio:
> Do we need a test in this case?

I can extract the coverage data that causes the crash in Vitest and add it as test case here.

SimenB (approved):

SimenB:
Thanks for following up @AriPerkkio! (and in general the help provided)

PR #804, opened 2024-10-25
fix: The color of 'missing if branch' is incorrect

travzhang:
[view case](https://github.com/canyon-project/canyon/tree/main/examples/vanilla-nyc)

3091729845741_ pic


bcoe:
@zhangtao25 this looks reasonable to me. What colour are you seeing in the report and what colour do you expect to see?

Showing a screenshot of before and after in the PR would be helpful.

SimenB:
image

this is what I see with the current release - the `E` seems more readable in that one? Or am I missing something?

PR #820, opened 2025-05-03
Revert "fix(deps): update make-dir CVE-2022-25883"

lpinca:
This reverts commit eab82a9aeff140a8fd2981c7f872830c985d479f.

`nyc@15` claims compatibly with Node.js 8 but eab82a9a broke it.

After the revert, the referenced security issue is still addressed because
`make-dir@3` depends on [`semver@6`](https://github.com/sindresorhus/make-dir/blob/6d029fe1/package.json#L45)
and the issue was fixed in `semver@6.3.1`.


bcoe:
@lpinca I'm hesitant to downgrade a dependency, in support of a version of nyc 2 majors behind, in support of a Node.js version 16 majors behind, in support of fixing a CVE that isn't a true vulnerability to test a test coverage tool.

That's the motivation for this? i.e., is there a project that's running into problems that can't just upgrade nyc.

lpinca:
> That's the motivation for this? i.e., is there a project that's running into problems that can't just upgrade nyc.

"Yes", the `ws@7` release line which supports Node.js 8. It receives only security fixes, but at the moment tests fails due to this issue. https://github.com/istanbuljs/istanbuljs/commit/eab82a9aeff140a8fd2981c7f872830c985d479f was a breaking change. The revert is painless. I think it makes sense to revert and release `istanbul-lib-report@3.0.2`, then revert again and release `istanbul-lib-report@4.0.0`.

lpinca:
It basically did not change anything between `make-dir@3.1.0` and `make-dir@4.0.0`, see https://github.com/sindresorhus/make-dir/releases/tag/v4.0.0, and v4.0.0 was the only `make-dir@4` release, so reverting only once and releasing only `istanbul-lib-report@3.0.2` is also ok in my opinion.

lpinca:
If this is not acceptable I have other options like:

1. Keep CI broken
2. Exclude Node.js 8 from the matrix
3. Not getting coverage when testing on Node.js 8
4. etc.

Anyway, I think this is the correct fix for all projects that directly or indirectly depend on `istanbul-lib-report@3`, and I see no downsides.

PR #833, opened 2025-11-17
chore: Upgrades js-yaml to v4

pan-kot:
Update: the 3.13.2 of js-yaml also fixes the security violation, so feel free to merge or cancel the PR.

------

Upgrading js-yaml to resolve Dependabot warning:

## Impact
In js-yaml 4.1.0 and below, it's possible for an attacker to modify the prototype of the result of a parsed yaml document via prototype pollution (__proto__). All users who parse untrusted yaml documents may be impacted.

## Patches
Problem is patched in js-yaml 4.1.1.

## Workarounds
You can protect against this kind of attack on the server by using node --disable-proto=delete or deno (in Deno, pollution protection is on by default).

## References
https://cheatsheetseries.owasp.org/cheatsheets/Prototype_Pollution_Prevention_Cheat_Sheet.html


packages/istanbul-lib-instrument/test/specs.test.js
```diff
@@ -30,7 +30,7 @@ function loadDocs() {
         const filePath = path.resolve(dir, f);
         const contents = fs.readFileSync(filePath, 'utf8');
         try {
-            yaml.safeLoadAll(contents, obj => {
+            yaml.loadAll(contents, obj => {
```
pan-kot, line 33: See changelog: https://github.com/nodeca/js-yaml/blob/master/CHANGELOG.md#400---2021-01-03

Alex-ley-scrub:
the js-yaml issue is fixed in 4.1.1 but has also been backported and fixed in 3.14.2

https://github.com/advisories/GHSA-mh29-5h37-fv8m

PR #837, opened 2025-12-26
fix: `istanbul-lib-source-maps` to map functions without closing brace mapping

sapphi-red:
This PR fixes an issue in `istanbul-lib-source-maps` where functions were not mapped correctly when no closing brace mapping was present.

Oxc does not generate a source map entry for `}` ([sourcemap visualizer](https://evanw.github.io/source-map-visualization/#MTAwAGV4cG9ydCBmdW5jdGlvbiBpc0V2ZW4oYSkgewoJcmV0dXJuIGEgJSAyID09PSAwOwp9CmV4cG9ydCBmdW5jdGlvbiBpc09kZChhKSB7CglyZXR1cm4gIWlzRXZlbihhKTsKfQozMDMAeyJ2ZXJzaW9uIjozLCJuYW1lcyI6W10sInNvdXJjZXMiOlsidGVzdC50c3giXSwic291cmNlc0NvbnRlbnQiOlsiZXhwb3J0IGZ1bmN0aW9uIGlzRXZlbihhOiBudW1iZXIpIHtcbiAgcmV0dXJuIGEgJSAyID09PSAwXG59XG5cbmV4cG9ydCBmdW5jdGlvbiBpc09kZChhOiBudW1iZXIpIHtcbiAgcmV0dXJuICFpc0V2ZW4oYSlcbn0iXSwibWFwcGluZ3MiOiJBQUFBLE9BQU8sU0FBUyxPQUFPLEdBQVc7QUFDaEMsUUFBTyxJQUFJLE1BQU07O0FBR25CLE9BQU8sU0FBUyxNQUFNLEdBQVc7QUFDL0IsUUFBTyxDQUFDLE9BQU8sRUFBRSJ9)) because it does not affect the debugger nor stack traces. However, this caused Vitest's coverage feature to behave incorrectly when switching to Oxc (from esbuild). The problem was in the fallback logic that is used when a closing brace mapping does not exist. While it searched for other mappings on the same line, it did not search for mappings on different lines.

Apparently, a similar issue was found in esbuild in the past (https://github.com/evanw/esbuild/issues/1448) and was worked around on the esbuild side (https://github.com/evanw/esbuild/commit/c4d45e6b5766b1a1b56bf24b7dadb240de91332b). This PR would also allow those changes to be reverted.

### DIFF
--- a/packages/istanbul-lib-source-maps/lib/get-mapping.js
+++ b/packages/istanbul-lib-source-maps/lib/get-mapping.js
@@ -36,13 +36,27 @@ function originalEndPositionFor(sourceMap, generatedEnd) {
     // generated file end location. Note however that this position on its
     // own is not useful because it is the position of the _start_ of the range
     // on the original file, and we want the _end_ of the range.
-    const beforeEndMapping = originalPositionTryBoth(
+    let beforeEndMapping = originalPositionTryBoth(
         sourceMap,
         generatedEnd.line,
         generatedEnd.column - 1
     );
     if (beforeEndMapping.source === null) {
-        return null;
+        // search the previous lines as the mapping was not found on the same line
+        for (
+            let line = generatedEnd.line;
+            line > 0 && beforeEndMapping.source === null;
+            line--
+        ) {
+            beforeEndMapping = originalPositionTryBoth(
+                sourceMap,
+                line,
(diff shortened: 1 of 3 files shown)


AriPerkkio (approved):
All `@vitest/coverage-istanbul` tests pass with these changes ✅

PR #844, opened 2026-04-20
fix: infer names for class/object methods and assigned functions

### LINKED ISSUE #843: Object method names are reported as "(anonymous_XX)"
With the following js code (source.js):
```js
// Class with shorthand methods
class MyClass {
  regularMethod() {
    return 1;
  }

  methodWithDefaultParam(x = 42) {
    return x;
  }

  static staticMethod() {
    return 2;
  }

  get accessor() {
    return this._value;
  }
}

// Object literal with shorthand methods
const myObject = {
  objectMethod() {
    return 3;
  },

  // Anonymous function expression as property value
  funcExpression: function () {
    return 4;
  },

  // Arrow function as property value
  arrowMethod: () => {
    return 5;
  },
};

// Member assignment with anonymous function expression
myObject.assignedMethod = function () {
  return 6;
};

// Top-level named function declaration (should work fine)
function topLevelFunction() {
  return 7;
}
```

and the code for instrumenting it (plop.mjs):
```js
import { createInstrumenter } from "istanbul-lib-instrument";
import fs from "fs";
import path from "path";

const source = fs.readFileSync(
  path.join(import.meta.dirname, "source.js"),
  "utf8"
);

const instrumenter = createInstrumenter({ esModules: true });
instrumenter.instrumentSync(source, path.join(import.meta.dirname, "source.js"));
const { fnMap } = instrumenter.lastFileCoverage();

let bugCount = 0;
console.log("Function names recorded by istanbul-lib-instrument:\n");
for (const [id, entry] of Object.entries(fnMap)) {
  const line = entry.decl?.start?.line ?? "?";
  const isAnon = entry.name?.startsWith("(anonymous");
  if (isAnon) {
    bugCount++;
  }
  const marker = isAnon ? " ← BUG: should have a real name" : "";
  console.log(
    `  [${id}] line ${String(line).padEnd(3)} name: "${entry.name}"${marker}`
  );
}

console.log(
  `\n${bugCount} of ${Object.keys(fnMap).length} functions incorrectly named.`
);
```

running `node repro.mjs` leads to:
```
Function names recorded by istanbul-lib-instrument:

  [0] line 3   name: "(anonymous_0)" ← BUG: should have a real name
  [1] line 7   name: "(anonymous_1)" ← BUG: should have a real name
  [2] line 11  name: "(anonymous_2)" ← BUG: should have a real name
  [3] line 15  name: "(anonymous_3)" ← BUG: should have a real name
  [4] line 22  name: "(anonymous_4)" ← BUG: should have a real name
  [5] line 27  name: "(anonymous_5)" ← BUG: should have a real name
  [6] line 32  name: "(anonymous_6)" ← BUG: should have a real name
  [7] line 38  name: "(anonymous_7)" ← BUG: should have a real name
  [8] line 43  name: "topLevelFunction"

8 of 9 functions incorrectly named.
```

calixteman:
Class methods, object shorthand methods, function expressions used as object property values, and functions assigned to member expressions were all reported as "(anonymous_N)" in fnMap. Only named function declarations were correctly identified.

Introduce getFunctionName() which checks node.key for ClassMethod and ObjectMethod nodes, parent.key for ObjectProperty values, and parent.left.property for AssignmentExpression targets, falling back to the existing node.id / node.name logic.

Fixes https://github.com/istanbuljs/istanbuljs/issues/843


packages/istanbul-lib-instrument/test/varia.test.js
```diff
@@ -270,4 +270,37 @@ describe('varia', () => {
...
+            `class MyClass {
+                regularMethod() {}
+                static staticMethod() {}
+                get accessor() { return 1; }
+            }
+            const obj = {
+                objectMethod() {},
+                funcExpression: function() {},
+                arrowMethod: () => {},
+            };
```
AriPerkkio, line 289:
```suggestion
            `class MyClass {
                regularMethod() {}
                static staticMethod() {}
                get accessor() { return 1; }
                identicalName() {}
            }
            const obj = {
                objectMethod() {},
                funcExpression: function() {},
                arrowMethod: () => {},
                identicalName() {},
            };
```

This should produce `identicalName` and `identicalName_2` in `fnMap`.

AriPerkkio (commented):
Note that mapping function names into `fnMap` may break cobertura reporter integrations when multiple class constructors are in same file:
- https://github.com/AriPerkkio/ast-v8-to-istanbul/issues/169

packages/istanbul-lib-instrument/src/visitor.js
```diff
@@ -377,7 +407,7 @@ class VisitState {
             };
         }
 
-        const name = path.node.id ? path.node.id.name : path.node.name;
+        const name = getFunctionName(path);
```
nicolo-ribaudo, line 410: There is a `getFunctionName` function exported by `@babel/types`, would it be useful here?

PR #848, opened 2026-07-02
feat: support `skipFull` in `html` reporter

### LINKED ISSUE #609: Allow configuring skipFull for html reporter
The text reporter [allows for configuring `skipEmpty` and `skipFull`](https://github.com/istanbuljs/istanbuljs/blob/c1559005b3bb318da01f505740adb0e782aaf14e/packages/istanbul-reports/lib/text/index.js#L236-L237), but the html reporter [only allows `skipEmpty`](https://github.com/istanbuljs/istanbuljs/blob/c1559005b3bb318da01f505740adb0e782aaf14e/packages/istanbul-reports/lib/html/index.js#L261). Thus, running `nyc --reporter=html --skip-full` does not remove 100% covered files, which makes things very difficult to navigate when working on larger projects and thousands of folders and files.

This behavior is detailed in [an issue in the `nyc` repo](https://github.com/istanbuljs/nyc/issues/1020#issuecomment-636349267).

Would this project be open for adding this functionality? If so, I'll open a PR with the changes.

sparrowt:
Fixes #609

Support `skipFull` in the `html` reporter, in the same way as it's already supported in the `text` reporter, by omitting rows from the table for 100% covered files.

Note: I'm unsure how this repo manages cross-package changes, e.g. this relies on `CoverageSummary` gaining an `isFull` method in `istanbul-lib-coverage`, so how is this managed in terms of how a newer version of this will be required by `istanbul-reports`?

If this causes too much hassle, there's a simpler version here which avoids it by confining the changes to be purely within `istanbul-reports` https://github.com/istanbuljs/istanbuljs/pull/849


sparrowt:
Note: CI seems to be sad, see https://github.com/istanbuljs/istanbuljs/issues/850

PR #849, opened 2026-07-02
feat: support `skipFull` in `html` reporter (simplified version)

### LINKED ISSUE #609: Allow configuring skipFull for html reporter

sparrowt:
Fixes #609

> [!NOTE]
> This is an alternative fix which only touches `istanbul-reports`, in case https://github.com/istanbuljs/istanbuljs/pull/848 is too faffy due to cross-package dependencies.

Support `skipFull` in the `html` reporter, in the same way as it's already supported in the `text` reporter, by omitting rows from the table for 100% covered files.


sparrowt:
Note: CI seems to be sad, see https://github.com/istanbuljs/istanbuljs/issues/850

PR #851, opened 2026-07-15
fix: get CI passing on node 16 by pinning `node-releases`

sparrowt:
Pin node-releases to a version that does not require node>=18

Otherwise CI fails on node 16 due to the `engines` restriction added in https://github.com/chicoxyzzy/node-releases/commit/21c100c5aed67336fe43fc8b7f87fc6d73769808

Fixes #850

Issue #850, opened 2026-07-02
CI failing on node 16


sparrowt:
CI is failing e.g. here on my PR https://github.com/istanbuljs/istanbuljs/actions/runs/28598266340/job/84799328313?pr=849 and on another unrelated one too https://github.com/istanbuljs/istanbuljs/actions/runs/27890137448/job/82532113425
```
Run npm install --production --engine-strict --ignore-scripts --no-package-lock
npm WARN config production Use `--omit=dev` instead.
npm ERR! code EBADENGINE
npm ERR! engine Unsupported engine
npm ERR! engine Not compatible with your version of node/npm: node-releases@2.0.50
npm ERR! notsup Not compatible with your version of node/npm: node-releases@2.0.50
npm ERR! notsup Required: {"node":">=18"}
npm ERR! notsup Actual:   {"npm":"8.19.4","node":"v16.20.2"}

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/runner/.npm/_logs/2026-07-02T14_35_10_729Z-debug-0.log
Error: Process completed with exit code 1.
```

It then cancels all the other node versions too, AFAICS due to GitHub Actions defaulting `fail-fast` to `true`.

I can't find `node-releases` referenced anywhere in the repo, is it [this](https://www.npmjs.com/package/node-releases)?

Is this something where something in GitHub Actions has increased its minimum version of node?

sparrowt:
Ah, or is it simply that there's no lock file in this repo, so perhaps one of the `devDependencies` brings in `node-releases` at an un-pinned version?

Yeah and since https://github.com/chicoxyzzy/node-releases/commit/21c100c5aed67336fe43fc8b7f87fc6d73769808 it now specifies `"node": ">=18"` under `"engines"` which seems to be what's causing this install faioure, given the `--engine-strict`

PR #852, opened 2026-07-29
fix(istanbul-lib-coverage): merge bT against the pre-merge branch map

Eljees:
`FileCoverage.merge()` throws `TypeError: Cannot read properties of undefined (reading 'locations')`
when both sides track branch truthiness (`bT`) and the merge renumbers the branch map.

The order of operations is the problem. `merge()` replaces `this.data.branchMap` with the merged map:

```js
[hits, map] = mergeProp(this.b, this.branchMap, other.b, other.branchMap, keyFromLocationsProp);
this.data.b = hits;
this.data.branchMap = map;      // <- branch keys are renumbered here
```

and then merges `bT` against that *already replaced* map, while `this.bT` still holds the pre-merge
keys — and asymmetrically, `other.branchMap` is still the original one:

```js
if (this.bT && other.bT) {
    [hits, map] = mergeProp(this.bT, this.branchMap, other.bT, other.branchMap, keyFromLocationsProp);
```

`mergeProp` does `const item = aMap[key]` for every key of `aHits`, so as soon as the merged map is
shorter than the original, `item` is `undefined` and `keyFromLocationsProp` reads `.locations` off it.

The map gets shorter whenever two branch entries share `locations[0]`: `keyFromLoc` maps them to the
same key and they collapse into one entry. That is reachable with real reports — the coverage files
attached to #1538, for instance, contain `if` branches whose second location is the placeholder
`{"start": {}, "end": {}}`.

The fix keeps a reference to the pre-merge branch map and resolves both `b` and `bT` against it, so
the two sides are keyed against maps of the same vintage.

### Before / after

New test `merges bT when merging renumbers the branch map` in
`packages/istanbul-lib-coverage/test/file.test.js` — two branches sharing `locations[0]`, `bT` on
both sides, asserting every `bT` key still names an existing branch:

```
before:  26 passing, 1 failing
           merges bT when merging renumbers the branch map:
           TypeError: Cannot read properties of undefined (reading 'locations')

after:   27 passing
         42 passing   (coverage-map, file, index, percent)
```

### One thing I deliberately left out

Four lines above, `if (this.bT && other.bT)` means that merging a report that tracks branch
truthiness into one that does not silently drops `bT` altogether. I confirmed that separately — it is
data loss rather than a crash, and fixing it is a semantic decision (treat a missing `bT` as zeros?
carry the other side's over?) rather than a bug fix, so I would rather raise it on its own than mix it
into this PR. Happy to open that as a follow-up if you want it.

AI-assisted (LLM used for drafting); the change and both runs above are mine.


Eljees:
The linux jobs die in the install step: npm install --production --engine-strict fails with EBADENGINE - node-releases@2.0.51 requires node >=18. That is the pre-existing CI breakage tracked in #850, and #851 already has a green fix for it; it is unrelated to this diff.

The windows job, which runs the full suite, is green here, and the before/after numbers in the description are from local runs of the package tests.

Eljees:
Ping - open since 29 July, no review yet.

To save you the triage: the red linux jobs are not from this change. `npm install --production --engine-strict` fails in the install step with EBADENGINE, because `node-releases@2.0.51` requires node >=18. That is the pre-existing CI breakage tracked in #850, and #851 proposes a fix for it. On this branch `test (16)` is the only real failure - the other linux jobs are cancelled matrix siblings of it - and the `windows` job, which runs the full suite, is green.

Happy to rebase or rework if there is something you would want changed.

Eljees:
Ping - still open since 29 July, no review yet. The failing linux job is the pre-existing CI issue tracked in #850 (fix proposed in #851), unrelated to this diff; the windows job, which runs the full suite, is green here.

Eljees:
The red `test (16)` here is not this branch. The job fails during `npm install`:

```
npm ERR! code EBADENGINE
npm ERR! engine Not compatible with your version of node/npm: node-releases@2.0.51
npm ERR! notsup Required: {"node":">=18"}
npm ERR! notsup Actual:   {"npm":"8.19.4","node":"v16.20.2"}
```

`node-releases` reaches the tree transitively through browserslist and its current version dropped Node 16, so the Node 16 leg of the matrix can no longer install dependencies. The other legs are cancelled by fail-fast rather than failing on their own.

It is repo-wide: `test (16)` fails on every pull request whose run is recent - #858, #856, #854, #848, #849 and this one - and passes only on the ones whose last run predates that release. `main` is green for the same reason, its run is older.

Dropping 16 from the matrix, or pinning `node-releases`, would clear it, but that is a supported-versions call rather than something to slip into this PR. Happy to send either as its own change if you tell me which you would prefer. #854 is stuck behind the same thing.

Eljees:
Following up - still open since 29 July with no review. To recap: the linux failures here are the pre-existing EBADENGINE issue from node-releases@2.0.51 (tracked in #850, fix proposed in #851), not this diff; the windows job, which runs the full suite, is green. Happy to rebase or adjust if there's feedback.

Eljees:
The red on this PR is the `test (16)` leg, and it is not from the diff — it fails the same way on untouched `main`. Writing it up once because it is the only thing standing between this PR and a green tick, and the same applies to #854.

`test (16)` never reaches a test. It dies in the install step, which the workflow runs as

```
npm install --production --engine-strict --ignore-scripts --no-package-lock
```

and `--engine-strict` refuses a transitive dependency that has dropped Node 16:

```
npm ERR! code EBADENGINE
npm ERR! engine Not compatible with your version of node/npm: node-releases@2.0.51
npm ERR! notsup Required: {"node":">=18"}
npm ERR! notsup Actual:   {"npm":"8.19.4","node":"v16.20.2"}
```

The other six matrix legs on this PR are `cancelled`, not failed — fail-fast stops them once `test (16)` goes red.

I checked it against the base rather than assuming. In a `node:16` container (node v16.20.2, npm 8.19.4) on a fresh clone of `main` at `28ffdbc3` — the very commit whose stored checks show `test (16)` as green — that exact command fails with the same `EBADENGINE`, now naming `node-releases@2.0.56`. The stored green predates the release that dropped Node 16, so it does not reflect what `main` would do today.

Nothing to change on this branch for it. Whether Node 16 should stay in the matrix is a call for the maintainers; if you want it kept, dropping `--engine-strict` from that step or pinning the dependency would both do it, and I am happy to send that as a separate PR if it would help.

PR #854, opened 2026-07-31
fix(istanbul-lib-source-maps): do not duplicate path segments for root-relative sources

### LINKED ISSUE #718: Coverage for Single-File Vue Components Not Found
## Expected Behavior
When running `nyc` with the `lcov` reporter, I expect there to be an HTML file for viewing the code coverage of my single-file Vue components (files with the `.vue` extension).

## Observed Behavior
The `index.html` page of the `lcov` reporter output shows the coverage of my component, but when clicking on it to get a deeper view, the file cannot be found. It's listing the file's path twice:
```
<rootDir>/src/components/src/components/Menu.vue
```
When it *should be*:
```
<rootDir>/src/components/Menu.vue
```

## Repository that Reproduces Issue
https://github.com/tenkiller/vue_nyc

## Forensic Information
**Operating System:** Mac OS Sierra 10.12.6
**Environment Information:** [output.txt](https://gist.github.com/tenkiller/116ba5143767b996c80d4dff40e717ce)

Eljees:
## The bug (istanbuljs/nyc#718)

With webpack / vue-loader style in-place instrumentation, the inline source map's `sources` entries are relative to the **project root**, while the coverage object is keyed by the **absolute path of the very same file**. `pathutils.relativeTo()` resolves relative sources against the directory of the covered file, so the mapped coverage ends up keyed by paths like:

```
<rootDir>/src/components/src/components/Menu.vue
```

Reporters then cannot find the file - exactly what istanbuljs/nyc#718 describes (reported 2017, still reproduced on nyc 17 in Feb 2025; the thread pinpointed `get-mapping.js` back in 2019).

## The fix

When a relative source matches the trailing path segments of the covered file's path, it refers to that same file - so return the covered file path instead of resolving against its directory. The comparison is per path segment (`file.js` does not match `file.min.js`), and sources containing `..` never take the shortcut.

Both call sites stay consistent: the coverage key in `get-mapping.js` and the `sourceStore` key in `map-store.js` share `relativeTo()`, so source content lookups keep matching the coverage keys.

## Runs

Red (test added first, before the fix):

```
AssertionError: expected [ Array(1) ] to deeply equal [ '/project/src/components/file.js' ]
-  "D:\\project\\src\\components\\src\\components\\file.js"
+  "/project/src/components/file.js"
```

Green after the fix: 11 passing, including a new control test asserting that a relative source naming a **different** file (`file.js` next to `file.min.js`) still resolves against the covered file directory exactly as before.

Fixes istanbuljs/nyc#718.

AI-assisted (LLM used for drafting); the analysis and runs are mine.


Eljees:
The linux matrix dies in the install step: `npm install --production --engine-strict` fails with EBADENGINE because `node-releases@2.0.51` requires node >=18, so `test (16)` fails and the remaining matrix entries are cancelled by fail-fast. That is the pre-existing CI breakage tracked in #850, with a green fix already open in #851; it is unrelated to this diff.

The windows job, which runs the full suite, is green here.

Eljees:
Ping - open since 31 July, no review yet. Happy to rebase if that helps.

Eljees:
The red on this PR is the `test (16)` leg, and it is not from the diff — it fails the same way on untouched `main`. Writing it up once because it is the only thing standing between this PR and a green tick, and the same applies to #852.

`test (16)` never reaches a test. It dies in the install step, which the workflow runs as

```
npm install --production --engine-strict --ignore-scripts --no-package-lock
```

and `--engine-strict` refuses a transitive dependency that has dropped Node 16:

```
npm ERR! code EBADENGINE
npm ERR! engine Not compatible with your version of node/npm: node-releases@2.0.51
npm ERR! notsup Required: {"node":">=18"}
npm ERR! notsup Actual:   {"npm":"8.19.4","node":"v16.20.2"}
```

The other six matrix legs on this PR are `cancelled`, not failed — fail-fast stops them once `test (16)` goes red.

I checked it against the base rather than assuming. In a `node:16` container (node v16.20.2, npm 8.19.4) on a fresh clone of `main` at `28ffdbc3` — the very commit whose stored checks show `test (16)` as green — that exact command fails with the same `EBADENGINE`, now naming `node-releases@2.0.56`. The stored green predates the release that dropped Node 16, so it does not reflect what `main` would do today.

Nothing to change on this branch for it. Whether Node 16 should stay in the matrix is a call for the maintainers; if you want it kept, dropping `--engine-strict` from that step or pinning the dependency would both do it, and I am happy to send that as a separate PR if it would help.

